PR #30, opened 2020-07-08
ViewModel Events and Filters

davesmith00000:
The usefulness of the ViewModel has always been in debate, but a recent change to the Start up data, that allows you access to that data at all times, made one of the main practical uses of the view model redundant (i.e. as a sort of cache for things calculated during startup).

The other issue was dubious data flows. For example, there is a `ViewportResize` event - clearly a view concern that one might need to store details from - that could only been picked up by the model, and could only be given to the viewmodel by storing it in the main model so that the view model could read it! Bonkers!

So the decision was: Ditch view models, or beef them up, and we've gone for the latter.

The view model now receives global events the same way model does, but this added a new concern. Processing events is not cost free, so the viewmodel has gone from being called once to being called for every single event event if the view model isn't interested.

To remedy this we've extended the concept of event filters found in the SubSystems into the other games areas as well (and made all the terminology more consistent in the process!).

Event filters filter and map events in one go, and the events for games, scenes and subsystems are all filtered before any processing (not previously the case, as it turns out) to reduce calls. There is an `EventFilters.Default` which basically replicates the old behaviour, and is a good starting point since anything else is an optimisation.


demos/snake/snake/src/snake/gamelogic/ModelLogic.scala
```diff
@@ -31,9 +31,11 @@ object ModelLogic {
 
   def update(gameTime: GameTime, dice: Dice, state: GameModel): GlobalEvent => Outcome[GameModel] = {
     case FrameTick if gameTime.running < state.lastUpdated + state.tickDelay =>
+      println(1)
       Outcome(state)
 
     case FrameTick =>
+      println(2)
```
hobnob, line 38: You probably don't want these here

demos/snake/snake/src/snake/gamelogic/ModelLogic.scala
```diff
@@ -31,9 +31,11 @@ object ModelLogic {
 
   def update(gameTime: GameTime, dice: Dice, state: GameModel): GlobalEvent => Outcome[GameModel] = {
     case FrameTick if gameTime.running < state.lastUpdated + state.tickDelay =>
+      println(1)
```
hobnob, line 34: Likewise, remove this line

demos/snake/snake/src/snake/gamelogic/ModelLogic.scala
```diff
@@ -52,6 +54,7 @@ object ModelLogic {
       }
 
     case gameEvent =>
+      println(3)
```
hobnob, line 57: And remove this line

PR #34, opened 2020-07-13
Groups and tiles

davesmith00000:
The can of worms I mentioned the other day @hobnob, was to do with having just-good-enough support for Tiled (normal flat 2D grids only), and also fixing a bug where Groups weren't being transformed correctly.

I believe this branch does all that, although in regard to the Group problem I've "fixed" it by removing rotation and scale (the problem is that they don't commute and we have no notion of stacking transformations yet), and raised an issue to look into it separately. It's fixable with some thought, but not something I want to get into at the moment.


davesmith00000:
Thanks @hobnob !

Issue #46, opened 2020-09-05
Add Dotty Support


davesmith00000:
Library issues so far:

- [x] Wart Remover - needed?
- [x] Circe (scala.js artefacts)
- [x] UPickle (drop?) (scala.js artefacts)
- [x] uTest (scala.js artefacts)
- [x] ScalaCheck - red herring? Am I using this in the core since I restructured? (scala.js artefacts)

davesmith00000:
Indigo is now (at the time of comment) built against Scala 3.0.0-M2 and 2.13.4. You can only use Indigo with sbt at present due to missing mill dependencies.

PR #47, opened 2020-09-05
WIP: Dotty

davesmith00000:
An open PR so we can see what's going on with dotty support. Isssue #46


davesmith00000:
Most of Indigo now compiles against Scala 3, major caveats:

- I’m still using Scala 2 syntax - this is Scala 2 code compiling under Scala 3's compiler
- Code and tests compile without error
- Scala.js is missing or silently ignoring all of the crucial features that actually allow me to test anything, apparently including the test runner!
- The Json libs I support either haven’t been released yet or I don’t know how to use them in the new world, so a few problems there… Might be able to keep this as Scala 2 and use dotty compat as a worst case scenario…

Scala 3 - RC1 is due out this side of Christmas. The plan currently is to just keep updating stuff as it and supporting libraries are released. Luckily the dependencies are few (Yay, my strategy paid off for once!). With a bit of luck we’ll have a Scala 3 release for Christmas or early next year.

Issue #40, opened 2020-08-05
InputField should emit an event on text change


davesmith00000:
Slight oversight in the creation of input fields. InputFields work fine and their values can be read at the view model or view level.

However, it is difficult to get information out of the InputField and back "upstream" into the model. Unfortunately that's probably one of the main use cases, e.g. "Enter your name" store it in the model.

The implementation of InputField does notice when a change has occurred, and it should be fairly straightforward to emit a "ChangeEvent" of some kind. It may be better to let the user supply a custom event to return on change, or else we'll need to add some sort of identifier to the InputField so that the user can understand which input field sent the message.

PR #51, opened 2020-09-07
Add #40: InputFields emit an event on text change

davesmith00000:
As described in issue #40.

The main API impact that causes a breaking change is that updating an InputField now returns an `Outcome[InputField]` instead of an `InputField` so that we can capture the event.

Events are only emitted if you've given the input field an ID/Key so that you can track which input field the event came from.

### DIFF
--- a/indigo/indigo-extras/src/test/scala/indigoextras/ui/InputFieldTests.scala
+++ b/indigo/indigo-extras/src/test/scala/indigoextras/ui/InputFieldTests.scala
@@ -14,6 +14,15 @@ import indigo.shared.datatypes.FontInfo
 import indigo.shared.AnimationsRegister
 import indigo.shared.time.GameTime
 import indigo.shared.datatypes.Point
+import indigo.shared.events.InputState
+import indigo.shared.dice.Dice
+import indigo.shared.FrameContext
+import indigo.shared.input.Mouse
+import indigo.shared.input.Keyboard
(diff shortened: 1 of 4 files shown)


hobnob (commented):
:+1: All looks good

hobnob (approved):

indigo/indigo-extras/src/test/scala/indigoextras/ui/InputFieldTests.scala
```diff
@@ -192,8 +201,54 @@ object InputFieldTests extends TestSuite {
...
+
+            val field =
+              InputField("", assets).giveFocus.withKey(BindingKey("test"))
+
+            val actual =
+              field.update(context).flatMap(_.update(context))
+
+            actual.state.text ==> "ABCABC"
+            actual.globalEvents.head ==> InputFieldChange(key, "ABC")
+            actual.globalEvents(1) ==> InputFieldChange(key, "ABCABC")
```
davesmith00000, line 218: For future reference. I look at this today and though.. does that mean they're in the wrong order? But as it turns out, events are pushed into the queue from head to tail, and the queue is a fifo queue:

https://github.com/PurpleKingdomGames/indigo/blob/009b2f3a3f88f6d2fb0631dd1378fe9ba6372446/indigo/indigo-core/src/main/scala/indigo/gameengine/GameLoop.scala#L70

PR #53, opened 2020-09-12
Bringing back case classes...

davesmith00000:
## Introduction

Some time ago in the ancient past, I banished case classes from the Indigo code base. I cannot now remember why I did this, other than a vague idea that I thought I was helping performance. I'm confident that I had no concrete evidence to base that decision on.

I've been wondering about the (lack of?) wisdom of that decision ever since, and after recent discussions with @drsimonmiles I have decided to get to the bottom of the matter in a _slightly but hardly conclusively_ scientific way.

## Hypothesis

1. In the worst case, case classes will be no less performant than classes. The additional methods case classes generate, and the persisted fields they have will have barely any impact on a fully optimised Indigo game.

2. In the best case, the implementation of case classes and their use of shallow cloning may have a positive impact on memory usage and garbage collection.

## Method

We will test three versions of Indigo against the Cursed Pirate Demo.

The cursed pirate demo is the only Indigo game that does any real work every frame without requiring user input, which could affect the results because I'd need to do the same kinds of actions over and over again fairly consistently.

The three version we will test are:

1. The current public release 0.3.0
2. The "master" branch, which has no case classes on hot code paths.
3. The "case-classes" branch, where most of the public API has been converted to case classes - as you can see for yourself in this Pull Request.

The public release is our baseline, but it's important to test the master branch since some changes have already been made there, and they may affect performance.

We will test each version three times, by:

1. Rebuilding the demo with `sbt clean compile fullOptJS indigoBuildFull`
2. Opening a Firefox window and its performance window.
3. Running the game for 30 seconds to let the JS engine warm up.
4. Taking a memory snapshot.
5. Recording the performance for 30 seconds, and then stopping.
6. Where we need to look at time series values, we will only sample from the second 10 seconds in recorded 30 second window, to reduce the effects of starting and stopping the profiler. I also picked frames at random from periods where the game seemed to be running nominally (i.e. not right next to a big GC pause...)

This was difficult to measure, particularly since the profiler has an observable effect on performance, but I looked at:

1. File size
2. Memory footprint
3. Average frame duration
4. Observed GC pattern

## Results

I ran each three times as described above, but actually found them to be reassuringly consistent between builds/runs. The values below are averages, and all taken from the same computer.

All of the big GC blocks followed a near identical pattern at regular intervals. Mostly NonIncremental GC, some CC graph reduction, and one-off Minor GC's and Cycle collections. The only thing that varied was the interval.

### Published version

Least stable frame rate, periodically bouncing between 59/60 FPS. Others were stable at 60.

- File size: 834KB
- Memory: 93.88MB
- Average frame duration: 8ms
- GC interval: 5 seconds

### Master

Smallest file size, _slightly_ better performance than master.

- File size: 820KB
- Memory: 93.36MB
- Average frame duration: 7ms
- GC interval: 5 seconds

### Case classes

Weirdly same file size as the published version. My hunch is that this has nothing to do with the code (which I checked was definitely the right version) and more to do with the closure compiler hitting some sort of complexity cut off. But I'm just guessing.

- File size: 834KB
- Memory: 93.46MB
- Average frame duration: 7ms
- GC interval: 7 seconds

## Conclusion

There's very little in it.

There appears to be no apparent performance difference between the current "master" and "case class" branches, and improvements over the published version seem due to commits that live on both branches.

The only value that has moved an any interesting way is the GC interval, where GC pauses are consistently further apart in the new version. This could actually be a bad thing, since in games the ideal would be to have a very regular, frequent, predictable, and small GC penalty. However it's hardly conclusive because it could have been doing the amount of clean up but simply not needing to do it as often. Certainly there was no observable frame duration penalty to suggest it was struggling.

What does this all mean?

Since there is no obvious performance reason not to use case classes, even if there's no performance improvement; Moving to case classes is still worth doing since it's cleans up the code base and presents more usable objects to the end user.

So if we're happy with the PR, we should go ahead and merge it.

### DIFF
--- a/examples/assetLoading/src/main/scala/com/example/assetloading/AssetLoadingExample.scala
+++ b/examples/assetLoading/src/main/scala/com/example/assetloading/AssetLoadingExample.scala
@@ -43,6 +43,7 @@ object AssetLoadingExample extends IndigoDemo[Unit, Unit, MyGameModel, MyViewMod
       }
     )
 
+  @SuppressWarnings(Array("org.wartremover.warts.ToString"))
   def updateModel(context: FrameContext[Unit], model: MyGameModel): GlobalEvent => Outcome[MyGameModel] = {
     case AssetBundleLoaderEvent.Started(key) =>
       println("Load started! " + key.toString())
--- a/examples/inputfield/src/main/scala/indigoexamples/InputFieldExample.scala
+++ b/examples/inputfield/src/main/scala/indigoexamples/InputFieldExample.scala
@@ -40,12 +40,10 @@ object InputFieldExample extends IndigoDemo[Unit, Unit, Unit, MyViewModel] {
 
   def updateViewModel(context: FrameContext[Unit], model: Unit, viewModel: MyViewModel): GlobalEvent => Outcome[MyViewModel] = {
     case FrameTick =>
-      Outcome(
-        viewModel.copy(
-          singleLine = viewModel.singleLine.update(context),
-          multiLine = viewModel.multiLine.update(context)
-        )
-      )
+      for {
+        single <- viewModel.singleLine.update(context)
+        multi  <- viewModel.multiLine.update(context)
+      } yield viewModel.copy(singleLine = single, multiLine = multi)
 
     case _ =>
       Outcome(viewModel)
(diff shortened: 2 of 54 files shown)


davesmith00000:
There are some other changes I made along the way here, fixing / making terminology consistent, syntax corrections, adding missing methods (again for consistency), renaming a few ugly things, mostly around `Flip`.

Issue #52, opened 2020-09-09
TextureAtlas Loading image data is insecure


laurensdv:
Hello,

after loading succesfully images in the indigo demo application:
`[Log] [INFO] [Indigo] [Image] Loading assets/dots.png (mygame-indigo_demo-fastopt.js, line 75157)`
`[Log] [INFO] [Indigo] [Image] Success assets/dots.png (mygame-indigo_demo-fastopt.js, line 75157)`
`[Log] [INFO] [Indigo] Asset load complete (mygame-indigo_demo-fastopt.js, line 75157)`
`[Log] [INFO] [Indigo] Creating atlases. Max size: 4096x4096 (mygame-indigo_demo-fastopt.js, line 75157)`

I get the following error, right after the atlas creation instruction:
`Unable to get image data from canvas because the canvas has been tainted by cross-origin data.
(anonymous function) — TextureAtlas.scala:226
(anonymous function) — TextureAtlas.scala:248
(anonymous function) — TextureAtlas.scala:267:110
(anonymous function) — AnonFunctions.scala:27
create__sci_List__F1__F2__Lindigo_platform_assets_TextureAtlas — TextureAtlas.scala:38
createTextureAtlas__Lindigo_platform_assets_AssetCollection__s_util_Try — List.scala:245
initialise__s_util_Try — Platform.scala:39
(anonymous function) — GameEngine.scala:142
(anonymous function) — GameEngine.scala:107
foreach__F1__V — Try.scala:260
run__V — Promise.scala:444
scala$scalajs$concurrent$QueueExecutionContext$PromisesExecutionContext$$$anonfun$execute$2__jl_Void__jl_Runnable__sjs_js_$bar — QueueExecutionContext.scala:52
promiseReactionJob
$p_jl_JSConsoleBasedPrintStream__doWriteLine__T__V — System.scala:389`

See:
https://github.com/PurpleKingdomGames/indigo/blob/b3bb061b609f819358fd361e3bc58233c8d9c3be/indigo/indigo-platforms/src/main/scala/indigo/platform/assets/TextureAtlas.scala#L226

The issue occurs when trying to run it in an ios or android cordova container. My impression is that it has to do with recent restriction being put on loading files through file:/// protocol.

How could I resolve this issue? Is this something that will be addressed, or is it something that can be workaround, e.g. hosting the images on a CDN rather than including them in the app assets?

davesmith00000:
[...]

Issue #49, opened 2020-09-06
Fullscreen support


davesmith00000:
Indigo game devs can respond to resizing events by listening to the `ViewportResize` event. When the canvas is resized - if you do nothing - everything stays the same it's just that the canvas space is adjusted. Hence, when Indigo is put into fullscreen mode by an external script, everything works just fine.

What would be good is a way to trigger fullscreen mode from within a game. Perhaps a `FullscreenToggle` event.

davesmith00000:
There is an "interop.html" file in the sandbox directory that does fullscreen, but it looks quite limited, see below instead:

Documentation: https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen

Working example: https://fullscreen-requestfullscreen-demo.glitch.me/
HTML: view-source:https://fullscreen-requestfullscreen-demo.glitch.me/
JS: https://fullscreen-requestfullscreen-demo.glitch.me/script.js

I think it would be ok to add the start up code to the plugin template as a bit of "stuff you must have for this to work".

PR #63, opened 2020-09-25
Add Cordova support to Mill & SBT plugins

davesmith00000:
Off the back of the excellent work by @laurensdv in issue: #52

I have issues testing Cordova (boring personal computer problems) so I've only included the platforms I think I can test for now, i.e. Browser, Electron, iOS, and at a stretch OSX.

I'm also not a Cordova expert, the hope is to lay the foundations for Cordova support that hopefully others can improve on - or I will as my own needs arise.

Worth noting that I haven't finished testing this yet.


davesmith00000:
I have tested Browser, (Cordova's) Electron, and iOS and they all work fine.

I've removed OSX. Spent several hours trying to get OSX on Cordova to run WWW from a localhost (as we did with iOS) without success, and then it hit me: Why would you ever mess about with Cordova for Windows or OSX when you can just use Electron - which works perfectly! - either standalone (as in the indigoRun output folder) or from the Cordova build.

So the only real omission is Android. I haven't added Android as I'm having issues testing but hopefully someone else will do that at some point. Or I'll do it when I work out a build environment.

davesmith00000:
To use cordova here, you need to have cordova installed:

`npm install -g cordova`

and electron

`npm install -g electron`

Then do (with Mill, SBT very similar):

```
mill mygame.fastOpt
mill mygame.indigoCordovaBuild
```

Navigate to `out/indigoCordovaBuild/dest/` and do:

```
cordova platform add ios
cordova platform add browser
cordova platform add electron
cordova prepare
```

...and then hopefully one of:

```
cordova run ios
cordova run browser
cordova run electron
```

laurensdv:
Hello @davesmith00000 

The cordova build works for me as well.

Had to update sbt plugin to 1.2.0 and change all references now to 0.3.1-SNAPSHOT for indigo.
StartupErrors in setup; and Keys.SPACE (in favour of SPACE in Constants) seem deprecated.

Some remarks when building with sbt:

`<PROJECT HOME>/out/mygame/indigoCordovaBuild/dest/`
is
`<PROJECT HOME>/target/indigoCordovaBuild` (no dest folder)

Then `cordova build ios` puts everything in
`<PROJECT HOME>/platforms/ios` (somehow I was expecting something inside target, but it could be expected behaviour)

Not sure if it is required for stability for cordova apps that the indigo game launch should wait for the device ready event of Cordova.js or not... (https://cordova.apache.org/docs/en/latest/cordova/events/events.html)
See also: 
https://github.com/laurensdv/helloworldindigo/blob/f34312207aa41ac10e4a81b7b827f17daf8bb1a2/index-dev.html#L100

To enable the fullscreen (resize), I guess I have to configure something more in the GameConfig, viewport settings or in another place?
screenshot

laurensdv:
> Then `cordova build ios` puts everything in
> `<PROJECT HOME>/platforms/ios` (somehow I was expecting something inside target, but it could be expected behaviour)

Resolved by removing config.xml and package.json at <PROJECT_HOME>

> To enable the fullscreen (resize), I guess I have to configure something more in the GameConfig, viewport settings or in another place?

Note that the app seems to resize the canvas, but that the web gl content is still drawn in original size in the top left corner.

davesmith00000:
All you notes at the top look accurate, and reflect some changes in 0.3.1-SNAPSHOT (like startup errors and Keys).

> Then cordova build ios puts everything in `<PROJECT HOME>/platforms/ios`

I've been `cd`ing into the `target/indigoCordovaBuild` directory before doing all the `cordova prepare` type stuff.

> Not sure if it is required for stability for cordova apps that the indigo game launch should wait for the device ready event of Cordova.js or not... 

I was wondering about that... happy to be advised. I see this as a getting started point, not a definitive solution. If you (by which I mean anyone wanting to use Indigo with Cordova) want to publish via Cordova you are definitely going to have to roll up your sleeves, read the docs and work on your integration code I think. Although of course if there is just "a good way" to do all this stuff I'm delighted to get PRs that improve the basic set up.

Fullscreen has an open issue I'm planning to work on next. The ticket does point out that you don't need Indigo to do this, as such, in your embed code you can just tell the canvas to go into fullscreen mode. https://github.com/PurpleKingdomGames/indigo/issues/49

> Note that the app seems to resize the canvas, but that the web gl content is still drawn in original size in the top left corner.

I'll need to look into it further. My gut instinct is that it's probably doing the right thing based on what it knows at the time, but maybe as you said, we need to wait for the device ready event? Oh! Also, in this version by default we pass the height and width into the game as flags, but you need to actually use them, might help?

https://github.com/PurpleKingdomGames/indigo/blob/94560f3af1981b4faff3ed170b18a3aa9280440b/indigo/sandbox/src/main/scala/com/example/sandbox/SandboxGame.scala#L23

laurensdv:
On your last remark:

indeed this should be the case:

https://github.com/laurensdv/helloworldindigo/blob/f34312207aa41ac10e4a81b7b827f17daf8bb1a2/src/main/scala/config/MyGameConfig.scala#L13-L16

https://github.com/laurensdv/helloworldindigo/blob/f34312207aa41ac10e4a81b7b827f17daf8bb1a2/src/main/scala/MyGame.scala#L24-L27

However the screensize set in the viewport doesn't match the screensize of the iOS device.
I think it is a matter of expectation, I was expecting that it would resize to fullscreen with the canvas.

Not that in safari debug, the canvas is resize properly to the max of the screen including the black background, however, the viewport (or the content in the viewport?) is not resized with it.

screenshot

However I noted in the startup logs in the console:
ViewPoint [412, 896]

Not sure if this has anything to do with the display size of the elements within the canvas.

laurensdv:
Some additional remarks:

Replacing my code with the flags more explictly like your example has no impact

```scala
val gameViewport =
      (flags.get("width"), flags.get("height")) match {
        case (Some(w), Some(h)) =>
          GameViewport(w.toInt, h.toInt)

        case _ =>
          GameViewport(MyGameConfig.width, MyGameConfig.height)
      }
    BootResult
      .configOnly(MyGameConfig.config.withViewport(gameViewport))
      .withAssets(MyAssets.assets)
      .withAnimations(MyAssets.animations)
      .withFonts(MyAssets.fontInfo)
      .withSubSystems(Set(FPSCounter(MyAssets.fontKey, Point(5, 5), 60)))
```

I also noted, after in depth looking, that initially the canvas size is wrong,
then it is resized, then the that the first (couple of) frame(s), elements within the canvas are sized correctly before going back to the wrong size. 

See below slowed animation:

animation

davesmith00000:
Right, good stuff, let's go through it:

First, you're symlink is going to `target/indigoBuild` but it should go to `target/indigoCordovaBuild` in order to get the resizing behaviour and initial flags. The HTML/JS output is a little different between `indigoBuild` and `indigoCordovaBuild` (actually it's based on the HTML/JS generated for `indigoRun`).

Let's break down what's happening in that slowed down animation (which is very helpful by the way), I think what's happening is:

1. The game is initially loaded at the size dictated by the config - I don't think it's getting the flags?
2. Then the canvas is resized
3. Then when it renders "correctly" that is actually a glitch where the buffer is just being scaled up to meet the canvas before the renderer produces the next frame. I need to find a good work around for that. It could be done in JS/CSS I think (hide the element during resize and show it again once it's finished - or something).
4. Indigo catches up and renders it actually correctly i.e. no stretch/scale just a bigger window area.

The idea here was to resize the game without deforming graphics through arbitrary scaling. In your case you're scaling a game with the correct dimensions so the scale up looks ok, but if you had changed the aspect ratio it would be awful.

The make the game scale the way you want, you have two options:

1. Don't use the scaling method provided! You can get the effect you want with CSS (that was where I started by accident). :-)
2. The intention was that the developer - knowing the dimensions at boot time - could write code to render things correctly at different resolutions and aspect ratios - even render a different view entirely. In your specific case, I _think_ all you need is to calculate a new magnification and Indigo will do the rest, i.e. it looks twice as big, so multiply your magnification by 2 and see how it looks...

Is there also a use case for "scale to fit"? Maybe. We used to have that in the days of Flash but I didn't know a Flash dev worth anything who used it because it only looked ok in the most simplistic of cases. Often you'd see off screen elements being drawn because the dimensions of the animation didn't fit the aspect of the window it was being drawn in.

Here is the sandbox running Cordova iOS, not that it hasn't been scaled up either:

cordova_ios

Does that help?

laurensdv:
Yes it helps, certainly! I was expecting a full screen render, but it is just being rendered in a bigger canvas.

Some comments:

* Your remark about the symlink & flags. This old link is still in my repo, but I am not using it anymore. 

However, you are right about the flags not being taken into account. This was because I only did the `sbt indigoCordovaBuild`, but didn't recompile and build the project as a while using the `sbt buildGame` shortcut. 

Doing all the steps correctly results in, the same view, however this time the viewpoint is correctly set and I did a `println `on the flags just to be sure:

screenshot

* This approach is perfect when the screensize ( & rotation: landscape or portrait) of the target of the advice is known...
If not, then I guess you're right a CSS approach and tweaking the index.html is probably a better choice (bearing in mind your remark of the old flash developments)
I investigated this and came across a few 'boundary' cases, that may or may not be so boundary, depending on the platforms targeted (browser, mobile...)

https://www.sitepoint.com/modernize-your-html5-canvas-game/

If you check the following file, it should take some of these aspects into account, though it is not compatible with the flags approach.

https://github.com/laurensdv/helloworldindigo/blob/master/index-dev.html

which brings me to the next two points:

* where can I stage such a 'tweaked' html file or template to be used instead of the default?

* I noticed that when doing a `indigoCordovaBuild `the entire `target/indigoCordovaBuild` folder is being cleared and replaced. Also the config.xml etc. Is it possible to stage these files to override them in the target. Also think of splash screens or icons of the application, if I would configure them in the target directory, then a next build would clear all of it.

davesmith00000:
> https://www.sitepoint.com/modernize-your-html5-canvas-game/
Good link! Thanks for sharing.

I'm going to answer both your points at the same time.

I didn't want to get into the business of managing publishing, I only wanted to give people a basic reference (or a complicated one, PR's welcome! :-)). So at the moment it's doing the standard Mill / SBT thing of clearing the folder and rebuilding - which I think is reasonable since you can't ever rely on a folder in the target dir (assuming sbt) being there because one misjudged clean and it's gone.

But I fully take your point.

I think there's two ways forward (maybe both):

1. Build tool education.
2. Further improve the plugin.

Build tool education: What I mean is that arguably the best solution is to write a custom task/command/process that inspects the build folder and copies the parts you care about into your staging folder (outside the target dir). Better yet, it would merge your custom assets with selected artefacts from the build into a new folder that could be reproduced.

Which brings me to "Further improve the plugin": We could improve the plugin to do that for you. For example, you (the user) could define a path to a folder full of the publish assets (i.e. your custom config.xml and anything else in the right folder structure) which is then merged with the build output (may also need to be able to set some build layout rules or filters or something - could be tricky).

I'm tempted to see if we can convince everyone to do the former, i.e. custom build tasks, because although we _could_ make the plugin cleverer... I suspect it would never be clever enough. Eventually I suspect most builds are going to end up a bit unique somehow or other. I could be wrong, time will tell. I sort of feel like: If you've got as far as Cordova you should be afraid of a little file munging in a build - could even be an external shell script right? [Could even be a shell script you call from SBT! :-)](https://github.com/PurpleKingdomGames/indigo/blob/master/indigo/build.sbt#L217)

I would actually do the custom task thing in order to work out how to do the plugin improvement. If you decide to [have a go](https://www.scala-sbt.org/1.x/docs/Tasks.html), do yourself a favour and use [os-lib](https://github.com/lihaoyi/os-lib) if you can.

All that said, if you think there's value improving the publish experience via the plugins, you would like to describe how you'd like it to work in an issue? Happy to consider it. :-)

davesmith00000:
Just in case "Build tool education" is misinterpreted in any way, all I really meant is that I could write something in the docs for future developers with a similar problem, demonstrating how you could do such a thing. A lot of people find SBT (in particular) and Mill task writing a bit mysterious.

laurensdv:
Thank you for the clear explanation.

I must agree with you, it is a matter of clearly explaining - "build tool education" - to manage the expectations. Getting a game in an adequate resolution would just be the start of any deployment to a mobile or other platform.

I do like the idea of a folder to stage overrides of the deployment target, so when it gets replaced following a new built, the overrides are being merged with the new target.

In any case, it is good to know and to have tested, that with the right guidance & customizations deploying to multiple target platforms (using cordova) is possible.

PR #64, opened 2020-09-29
Radio Buttons - API revision

davesmith00000:
This is a follow up to the work by @drsimonmiles in PR https://github.com/PurpleKingdomGames/indigo/pull/60

After merging the new Radio Buttons component I started playing with the example. Everything worked perfectly, but I was very mildly irked by user experience of having to declare the same option twice to mark it as being initially set:

```scala
RadioButtonGroup(
  ...,
  options = List(option1, option2, option3),
  ...,
  selected = Some(option1)
)
```

...and thought it would be nice if you could do something like:

```scala
RadioButtonGroup(assets, 10, 10)
  .addRadioButton(option1.selected)
  .addRadioButton(option2)
  .addRadioButton(option3)
```

Pulling that seemingly small harmless thread ...lead to a fair bit of rejigging. :-)

The structure is the same, but the user experience is a bit different (I hope cleaner), and the guts of the update logic is a rewrite to avoid using an object equality test (because we don't store the selected anymore, so we can't do that). I also added some very basic tests for a measure of confidence.

### DIFF
--- a/indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala
+++ b/indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala
@@ -26,24 +26,24 @@ final case class RadioButton(
     onHoverOut: () => List[GlobalEvent],
     state: RadioButtonState
 ) {
-  def withSelectedAction(actions: GlobalEvent*): RadioButton =
-    withSelectedAction(actions.toList)
-  def withSelectedAction(actions: => List[GlobalEvent]): RadioButton =
+  def withSelectedActions(actions: GlobalEvent*): RadioButton =
+    withSelectedActions(actions.toList)
+  def withSelectedActions(actions: => List[GlobalEvent]): RadioButton =
     this.copy(onSelected = () => actions)
 
-  def withUnselectedAction(actions: GlobalEvent*): RadioButton =
-    withUnselectedAction(actions.toList)
-  def withUnselectedAction(actions: => List[GlobalEvent]): RadioButton =
+  def withUnselectedActions(actions: GlobalEvent*): RadioButton =
+    withUnselectedActions(actions.toList)
+  def withUnselectedActions(actions: => List[GlobalEvent]): RadioButton =
     this.copy(onUnselected = () => actions)
 
-  def withHoverOverAction(actions: GlobalEvent*): RadioButton =
-    withHoverOverAction(actions.toList)
-  def withHoverOverAction(actions: => List[GlobalEvent]): RadioButton =
+  def withHoverOverActions(actions: GlobalEvent*): RadioButton =
+    withHoverOverActions(actions.toList)
+  def withHoverOverActions(actions: => List[GlobalEvent]): RadioButton =
     this.copy(onHoverOver = () => actions)
 
-  def withHoverOutAction(actions: GlobalEvent*): RadioButton =
-    withHoverOutAction(actions.toList)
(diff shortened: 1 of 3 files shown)


davesmith00000:
Replaying a mildly abbreviated discord chat with @drsimonmiles here for posterity:

**Feedback:**

1. Does this new scheme allow multiple radio buttons to be selected on creation of the button group? If so, is that intended to be an allowable state or just a consequence of providing a convenient interface?
2. I notice you refer to the size of button 'hit area' in the ScalaDoc, but I couldn't see where position of hit area was defined? Does that mean it must always be in the top left of the button?
3. Personally, I find it a bit unintuitive to use Point to contain size data... Couldn't there be a separate Size type with width and height parameters? Not that important, I realise.

**Last commit**

1. You can now only have more than one buttons selected if you put them in that state via the full case class constructor. All other methods are validated. I think that's strict enough.
2. & 3. I've replaced `size: Point` with `hitArea: Rectangle` which, now that it's called hit area, makes all the other problems go away. The hit area position is a relative offset from the radio buttons position, and you can have sensible syntax like size and width and height and so on.
I've also added more helper methods and made the interface even easier to use.

indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala
```diff
@@ -16,134 +16,215 @@ import indigo.shared.scenegraph.Group
...
+  def withUnselectedAction(actions: GlobalEvent*): RadioButton =
+    withUnselectedAction(actions.toList)
+  def withUnselectedAction(actions: => List[GlobalEvent]): RadioButton =
+    this.copy(onUnselected = () => actions)
 
-  def withHoverOverAction(action: () => List[GlobalEvent]): RadioButton =
-    this.copy(onHoverOver = action)
+  def withHoverOverAction(actions: GlobalEvent*): RadioButton =
+    withHoverOverAction(actions.toList)
+  def withHoverOverAction(actions: => List[GlobalEvent]): RadioButton =
```
hobnob, line 41: I realise this is in keeping with the above method... but should this be `withHoverOverActions` (plural) as you're adding more than 1?

indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala
```diff
@@ -16,134 +16,215 @@ import indigo.shared.scenegraph.Group
...
+    this.copy(onHoverOver = () => actions)
 
-  def withHoverOutAction(action: () => List[GlobalEvent]): RadioButton =
-    this.copy(onHoverOut = action)
+  def withHoverOutAction(actions: GlobalEvent*): RadioButton =
+    withHoverOutAction(actions.toList)
+  def withHoverOutAction(actions: => List[GlobalEvent]): RadioButton =
```
hobnob, line 46: Likewise here, whould we pluralise?

hobnob (commented):
Looks good! 🙂 Not sure about the pluralisation of some of the API thoug... thoughts?

indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala
davesmith00000: Fixed - good spot - thanks.

indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala
davesmith00000: Fixed

PR #66, opened 2020-10-09
WIP: PoC - Render using two shaders - DO NOT MERGE

davesmith00000:
This is just for reference.

It's a PoC to see what sort of changes are needed to render with more than one shader - definitely not to be merged.


davesmith00000:
This was a PoC, the real version is now out in the world!

PR #70, opened 2020-10-25
Added progress bar UI element and interleaved computation with example

drsimonmiles:
Add a very simple progress bar UI element, a class to control a computation whose steps are interleaved with frame updates, and an example of using the two together.


davesmith00000:
Thanks for submitting this. 🙂

I'll review the details more carefully later when I have a bit more time.

The bit that sticks out like a sore thumb is that `InterleavedComputation` should not live in the ui package. You made it to work with / because of the progress bar, but actually they are completely separate. For example: You could do a long running computation in a `SubSystem` and emit and event on competition without ever drawing anything, and equally you could render a progress bar over an NPC's head to show how far through a fishing task they were but with no expensive computation underpinning it.

drsimonmiles:
> The bit that sticks out like a sore thumb is that `InterleavedComputation` should not live in the ui package. You made it to work with / because of the progress bar, but actually they are completely separate. For example: You could do a long running computation in a `SubSystem` and emit and event on competition without ever drawing anything, and equally you could render a progress bar over an NPC's head to show how far through a fishing task they were but with no expensive computation underpinning it.

Sure, that makes sense. As you say, they are completely independent things that happen to work together if you want them to. I'm not sure where to suggest InterleavedComputation should go though.

davesmith00000:
Quick scout about, there's a thing called `indigoextras.datatypes.UpdateList` which does similar work (it takes a list of updatable things and only updates some of them each frame based on a pattern to spread the load out). I'd argue that `datatypes` is also a terrible package name, it's more of a bucket of strange things, but at least they would be consistently poorly arranged. :-)

The only other thing in that package is `TimeVaryingValue`s which arguably belong in the `temporal` package? But if you moved `InterleavedComputation ` to the same location I can rearrange them all sometime in the future. Unless you're feeling inspired and want to move them all to something better named? Don't let me stop you!

drsimonmiles:
Sorry for the delays in replying - work's heavy... How about I move them to a package 'interleaved'? That would seem an OK category label for UpdateList and InterleavedComputation.

davesmith00000:
Go for it.

davesmith00000:
Thanks Simon - I'm going to merge it to a branch I think. Broadly it all looks great, there's a couple of bits I'd like to review more carefully before it hits master. I'll send over a PR.

PR #71, opened 2020-11-02
Progress bar - further PR review

davesmith00000:
This is a PR submitted by @drsimonmiles, there was a couple of bit of clean up I wanted to do as well as a more careful reading of the `InterleavedComputation`.


davesmith00000:
The low hanging clean up tasks are done and on this branch - nothing worth worrying about / reviewing.

I've moved onto the `InterleavedComputation` code. The code works, I've run the demo, the code looks neat and well written, and follows the conventions used in Indigo.

There is a couple of things I'd like to dig into further:

1. There are no tests - there might be good reasons for that but I'd like to try writing some to find out. Indigo is supposed to be testable. :-)
2. `System.currentTimeMillis` is used to time the operations, but this is a side effect, which may make testing difficult. Would like to try using `gameTime.delta` instead if possible.

davesmith00000:
Update on this PR.

This stuff looks easy enough but is actually quite fiddly.

Applying `LongCompute` to the progress bar example project appears to do the right thing if the tolerance is set to `0.05`, which seems very small. Anything bigger and the units to attempt just climbs, anything much smaller and it only ever does one unit at a time. This seems a really unintuitive number, should we expose it? Is it problem (factorisation in the example) specific? Is it just broken in some unexpected way?

drsimonmiles:
I've just tried LongCompute on my crowds simulation, where the computations are 'real' path searches. With the default tolerance of 0.05, the progress updates smoothly as with InterleavedComputation. Setting it to 0.1, the progress is glitchy with pauses and jumps. So I would conclude it is not a feature of the problem: the path search computations are individually much faster than the factorisation computations in the example.

If 0.05 always works, then I guess it is fine for it to be effectively hard-coded and hidden! But I don't know why that value would be the right one, so I'm not confident to say it is universal...

davesmith00000:
> If 0.05 always works, then I guess it is fine for it to be effectively hard-coded and hidden! But I don't know why that value would be the right one, so I'm not confident to say it is universal...

I feel the same. Thank you for testing it! When I get a bit more free head space I'm going to have one more go at it and if I fail, hardcode to 0.05 and merge it.

Question 1: `LongCompute` is supposed to be an improved version of `InterleavedComputation` to allow for testing. Do you agree that that's the case or do you feel `InterleavedComputation` does anything better?

Question 2: Which name do you like better? I'm on the fence. LongCompute is shorter, InterleavedComputation is more accurate. :-) (or alternatively, can you think of another name?)

drsimonmiles:
1. I'm probably biased by it being my own code, but I like that the original doesn't require any hidden quantities. It just takes the duration of the first step as the likely duration of future steps and adjusts later by averaging if that turns out to be inaccurate. If the client sets the relative size of the tasks appropriately, this should be a good approximation and give smooth interleaving in all cases, so it feels like the client has more straightforward control than adjusting LongCompute's tolerance. Clearly it is better to use GameTime.delta rather than currentTimeMillis for testability and probably other reasons, so if we could take that from LongCompute (and I don't see why we can't), that would be the ideal in my eyes.

2. In my mind, LongCompute better describes what it actually is as an artefact, while InterleavedComputation gives more clue as to how to use it. On the whole, I prefer LongCompute, i.e. better to be declarative (as well as short). The way it is used might be different in future versions of Indigo (e.g. maybe hiding the way it is called rather than the client having to do something each FrameTick), so InterleavedComputation might end up being a meaningless name while LongCompute would still be accurate.

When I have the brain capacity, I'm happy to try changing InterleavedComputation to use GameTime.delta.

davesmith00000:
Ok, if you're feeling inspired to give it a try I'm happy to merge that version instead of this one. There was some other clean up bits I did in this pr that we might need to recreate, but we can worry about that later.

The acid test is whether you can test it mind you. 😉

davesmith00000:
With regret, I'm going to close this PR down. It's a long way behind at this point, I still there's something here though.

PR #73, opened 2020-11-08
Group Transforms

davesmith00000:
Just after the initial launch, a bug was reported and I realised that group transforms didn't actually work, and had to remove rotation and scaling.

This branch is so far is just the ground work to start finding a solution.

The problem is quite simple: The scene graph is flattened into a simple list of display objects before being passed to the renderer. With movement you can just keep doing moveBy each time you flatten a group layer to end up in the correct place, and that works because you're manipulating the real position on the screen. But that doesn't work with rotation and scale, especially when you expect the rotation and scaling to be around the groups position. Display objects currently have no way to represent this.

Essentially you need to build up a stack of transforms and pass them to the GPU to be applied. An obvious structure for doing that is a Matrix, but this has three problems:

1. The Matrix4 implementation in Indigo is probably too slow, but ok, we could write a fast mutable version.
2. The current shader setup won't allow for any more data to be passed in (or not easily), could be fixed with a separate Uniform allocation though perhaps?
3. This does mean we're moving the matrix work off the GPU to the CPU, which would be a shame...


davesmith00000:
Note for later regarding sending the matrix to the vertex shader:

We currently send translation and scale combined as a vec4, with rotation, flipH and flipV packed into another vec4 alongside something else.

There's no actual need to send through a whole 4x4 matrix (4 x vec4), since we could send it in a compressed form. We only care about 6 of the float values (2D, yay!) from inside the matrix, in fact, which is one float less that we currently spend (if you include the flip values).

The matrix looks like:

```
t = translation
s = scale
r = rotate
s+r, r, 0, 0,
r, s+r, 0, 0,
0, 0, 0, 0,
t, t, 0, 0
```

davesmith00000:
Also the "send the matrix as a uniform" idea is a dud. Would result in a uniform set call per object which would be unacceptable in terms of performance I suspect. (could try it and check my assumptions of course.)

davesmith00000:
At the cost of a small performance penalty (expected, and actually not as bad as I'd feared it seems), DisplayObjects now encode all their transform data as a Matrix4.

The next task is that, while processing a group we need to somehow capture the groups transform matrix and apply it to each child node _after_ they've been turned into a display object...

davesmith00000:
The code is ready for review, I need to roll it out to the demos and examples, and update the group example accordingly.

PR #83, opened 2020-12-02
Input mapper example

maedjyuk-ghoti:
Move the dot around the screen with a top down view. No boundary checks.


davesmith00000 (approved):
This is great! Perfect minified example of how it works - thanks for putting it together! 👍

PR #88, opened 2020-12-17
Overhaul of Outcome Type

davesmith00000:
> This isn't ready to go yet!

### A. Removing global events from SceneUpdateFragment

All the functions return an `Outcome` except the ones that render things - but not any more! We now have:

```scala
def present(...): Outcome[SceneUpdateFragment]
```

This allows everything to compose more naturally and moves events out of SUF into the outer Outcome so that we could have error handling of scene generation code...

### B. Add error handling to Outcome

Indigo doesn't have concerns that require an IO-like type in general (no side-effects, no concurrency, single threaded etc.) but what it has been lacking is any reasonable error handling.

## Left to do...

1. Test error handling - prove the implementation I've concocted actually works!
2. Substitute in place of Outcome
3. Deprecate Outcome
4. Change the GameLoop logic to allow error recovery
5. Change the GameLoop logic to allow logging on crash


davesmith00000:
This work has taken on a life of it's own...

I completed the planned work and began to put together an error handling demo (which I still have not done, other than a skeletal project) when I realised that `Startup` allows you to emit events... and hey, it could fail too right? So really, I ought to move events out of `Startup` and wrap the whole thing in an `Outcome` ... and while you're doing that... why not do that to everything else is could conceivably involve processing that could fail / have interesting things to say with events? Like model initialisation?

To make all that work elegantly was actually quite easy... but involved replaced the much spurned but surely battle tested `Try` monad with `Outcome` - which is a bit scary, but consequently means that I _should_ get consistent error reporting that just flows through the engine (for the most part).

If any Cats fans are reading this, what I basically have in `Outcome` now, is a poor-man's `F: MonadError` (but not an effect type because there are no effects to speak of). Outcome _could_ just be an `F[(A, List[GlobalEvent])]: MonadError`. In the future, and say if we ever had threading to deal with then that might be interesting. At the moment though, Outcome is no more or less than the exact type needed, so requiring it seems more honest, easier to explain, and more usable.

The next round of work in this ever expanding line of work, is to put together the planned example and to test (and write tests) that show:

1. Everything can fail as expected in all new locations
2. Events are propagated as expected from all new locations

PR #104, opened 2020-12-29
QuadTrees...

davesmith00000:
The context here is that it is occasionally very handy to have a data structure that can help you search a cloud of points for an element.

I wrote a (probably very slow) quad tree implementation ages ago, and this PR brings it into Indigo, albeit stripped down and cleaned up. It's probably still slow (algorithm whizz kids can submit a PR), but it's faster than traversing a massive list, and lets you search by vertex, line, or bounding box.

To prove it works, I've removed the QuadTree implementation from Snake, and Snake now uses the Indigo version.

... while I was there, I've given Snake a long overdue code clean up. I wrote snake probably a couple of years ago and this is the first time I've really revisited it other than to superficially keep it inline with Indigo updates. The result is a much simpler code base that far easier to understand and probably a good bit more efficient.


davesmith00000:
Re-reviewed this morning. Rebuilt everything, applied to the changes to some local project. Seems to be fine.

PR #107, opened 2021-01-02
WIP: Tweening

davesmith00000:
After a few false starts, I think I've got a simple animation system that should work. Might need a better API but structurally it's feeling better.

However, this work is now blocked by the work on the group-transforms branch because work has been done there to standardise which transforms are available at the various levels of the SceneGraphNode hierarchy.


davesmith00000:
Over a year old so I'm closing this down I'm afraid. I suspect we'll come back to needing this though, or something very similar, because organising complex (not even all that complex!) animations is still difficult.

PR #116, opened 2021-03-01
Update Scala to 2.13.5

aumann:
I hope the scalafix/semanticdb change is Ok with you - I thought i'd bring it up-to-date with how it's documented here for sbt 1.3+ https://scalacenter.github.io/scalafix/docs/users/installation.html
I've put scala-2.13.5 and Scala.js into the same commit as scala-2.13.5 is not possible without the new Scala.js

I'll create a separate pull request for 3.0.0-RC1


davesmith00000:
Looks good to me!

davesmith00000 (approved):

PR #138, opened 2021-04-25
Fix overlapping rectanges with negative size

kag0:
I think this should work for `Rectangle`s with negative sizes, since getting the corners adds the size to the position.

If the other algorithm is preferred, I think it would work if we took some absolute values somewhere in the right hand side of the comparisons

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/datatypes/Rectangle.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/datatypes/Rectangle.scala
@@ -130,6 +130,9 @@ object Rectangle {
     b.x >= a.x && b.y >= a.y && (b.width + (b.x - a.x)) <= a.width && (b.height + (b.y - a.y)) <= a.height
 
   def overlapping(a: Rectangle, b: Rectangle): Boolean =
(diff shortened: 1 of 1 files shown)


davesmith00000:
Thank you kindly!

I need to have a think about this one. The process used to test the overlap is known as an Axis Aligned Bounding Box (AABB) ...but I bet I don't have a test case for negatively sized boxes.

The AABB approach is used in a few places so I need to have a dig around and see what else is affected - `BoundingBox` springs to mind for instance.

Another approach - I guess - would be to reposition the Rectangle so that top left was the actual top left of the area the rectangle covers. Wonder how other people cope with this? 🤔

kag0:
A short term solution would be to forbid negatively sized rectangles. Either gently through error logs, forcefully through crashing, or strongly through some typed whole numbers.

Then you could relax that restriction when you have a solution for comparing negative sizes

davesmith00000:
The problem is actually even more deeply rooted than I thought, I'm looking into it. The temptation to use something like refined to force positive numbers is massive, but that's not very friendly to newbies is it? Not to worry, I'm sure it can be fixed. :-)

PR #139, opened 2021-04-25
Log error on negatively sized rectangles

kag0:
rel https://github.com/PurpleKingdomGames/indigo/pull/138

in case rectangle overlap for negative sizes isn't fixed right away, an error message for the user to help them debug

negative sizes probably affects more than just overlap, but if that's not the case we could move this snippet into the overlap method


davesmith00000:
Closing as not needed. This was resolved in PR: https://github.com/PurpleKingdomGames/indigo/pull/142

PR #140, opened 2021-04-27
Scala 3 or Bust

davesmith00000:
On this branch:

1. I'm tracking the latest version of Scala 3
2. I've deprecated Scala 2
3. I'm converting the Scala 2 syntax to Scala 3*

*The aim of converting to Scala 3 syntax is to improve type checking, hopefully improve performance (with things like opaque types rather than AnyVal's), and so that things like multiversal equality will work in peoples games, i.e. relevant CanEqual instances will exist and so on.


davesmith00000:
Update, all* instances of `AnyVal` are now opaque types.

(* Except Outcome related extension methods that I'd like to think about.)

PR #148, opened 2021-05-05
Fonts & TextBox

davesmith00000:
Adds support for a new primitive type `TextBox` and a new `AssetType.Font`.

This implementation has some not immediately obvious limitations, such as.

- It's probably not incredibly performant.
- You can't chose a custom shader.
- You'd need to hand roll multiline support
- It's at the mercy of however the underlying browser renders the text.
- Not tested on all platforms
- No way to disable antialising

But this is good enough for a preview release, complementary to the existing `Text` primitive, which at the moment I see as complementary. `TextBox` is much much quicker to get started with, but `Text` would give you better results if you put in the work.


davesmith00000:
Things to review off the back of the original PoC:

- [ ] Add some other basic things to the scene so see if they're impacted.
- [ ] Need to force context change if it isn't happening.
- [ ] Making a whole other updateData function - ok?
- [ ] The gl params used to account for none power 2 textures - do the conflict, can we set them globally?
- [ ] Any other gl stuff I can not set every object?
- [ ] Can textContext be passed to LayerRender via constructor? And maybe not saved?
- [ ] Add more params to control the text.
- [ ] Add in a font asset type - Load or write it into the page.
- [ ] Can we get the size back out somehow?
- [ ] Cache the renderer textures?

davesmith00000:
The Good, The Bad, and The Ugly:

Good - Lots of options of rendering text!
Bad - Doesn't natively support wrapping text... (has to be implemented manually...)
Ugly - measuring text is going to be a pain (wrapping / reporting size).

davesmith00000:
(Copied in from a Slack discussion for posterity)
Font smoothing in browsers is a mess, so we're stuck with smooth fonts..

On the other hand, I think the problems with the new `TextBox` type actually just mean that the users now have choices and can make tradeoffs.
The easiest text thing to work with is actually good old `Text`, it's just a total pain to set up.
The fastest thing to use, is the new `TextBox` but it has limitations:
- No multiline
- No text measurements / boundary detection
- Font smoothing looks terrible in pixel art but would be fine for an HD UI

Issue #153, opened 2021-05-15
Introduce a "Spatial" typeclass


davesmith00000:
There is a boring problem at the moment that you can't do something like:

```scala
def doWork[T <: RenderNode](node: T): T =
  node.moveTo(Point(10, 10))
```

To solve this currently, you have to match on every possible concrete type you support - yuk.

It would be nice to introduce a typeclass to allow something like this:

```scala
def doWork[T : Spatial](node: T): T =
  node.moveTo(Point(10, 10))
```

bilki:
So, is this basically turning https://github.com/PurpleKingdomGames/indigo/blob/master/indigo/indigo/src/main/scala/indigo/shared/scenegraph/SpatialModifiers.scala into a typeclass, and give instances for every class extending from it? :thinking: Seems like a sensible approach, many (if not all) the implementations are there, it's a matter of moving them a proper place (companion object, instances object for all of them?) and fix the affected signatures by this change.

davesmith00000:
That's pretty much the idea yes. That trait is literally only there to make sure I don't forget any implementations! The only thing to be aware of follow recent work, is the SceneNode splits into RenderNode, EntityNode and DependentNode (CompositeNode is gone) and they have some slight differences.

I also wondered if there is another typeclass that allows you to also affect the attributes common the SceneNode's and Clone classes.

davesmith00000:
@bilki made a valiant attempt at this but we decided in the end that it created as many problems as it solved, and thought it best to leave it as is for now.

PR #154, opened 2021-05-19
Spatial typeclass #153

bilki:
It looks like too many changes but it's just moving code from here to there. Added a couple typeclasses named `BasicSpatial` and `Spatial` that replace `BasicSpatialModifiers` and `SpatialModifiers` classes. I had to implement a bit of autoderivation trickery for `Shape`, to avoid implementing its instances manually and match all the different types extending from it...

The instances implementation is almost the same across all the affected classes, so I guess there could be a way to refactor and provide a default implementation for many methods.

Also, tried to remove `withDepth` from some classes but it's required up in the `SceneNode` hierarchy, so I was unable to do it.


davesmith00000:
So `withDef` is only defined on the `SceneNode` because _sometimes_ what we really need is a `HasDepth` typeclass...?

davesmith00000:
The proof of the pudding is to push this out into places like `Button`, `Automata` (bit different) and radio buttons here:
https://github.com/PurpleKingdomGames/indigo/blob/252fc60a6ad7ae7b8d18073182669df44e187bef/indigo/indigo-extras/src/main/scala/indigoextras/ui/RadioButtonGroup.scala#L358

The idea - I think - would be that instead of taking a `SceneNode` and having to do an unreliable pattern match, you'd just take maybe a `T : BasicSpatial` and use that. That way you get exhaustivity.

indigo/indigo/src/main/scala/indigo/shared/scenegraph/Graphic.scala
```diff
@@ -51,53 +53,9 @@ final case class Graphic(
...
-  def flipVertical(isFlipped: Boolean): Graphic =
-    this.copy(flip = flip.withVerticalFlip(isFlipped))
-  def withFlip(newFlip: Flip): Graphic =
-    this.copy(flip = newFlip)
-
-  def withRef(newRef: Point): Graphic =
-    this.copy(ref = newRef)
-  def withRef(x: Int, y: Int): Graphic =
-    withRef(Point(x, y))
-
```
davesmith00000, line 100: This all looks great, but I'm surprised there's no reference to the Sandbox or Perf projects in the PR. Are they just working magically because they're implemented as extension methods?

indigo/indigo/src/main/scala/indigo/shared/scenegraph/Graphic.scala
```diff
@@ -145,4 +103,54 @@ object Graphic {
...
+      def withRotation(newRotation: Radians): Graphic =
+        graphic.copy(rotation = newRotation)
+      def withScale(newScale: Vector2): Graphic =
+        graphic.copy(scale = newScale)
+      def withDepth(newDepth: Depth): Graphic =
+        graphic.copy(depth = newDepth)
+      def withFlip(newFlip: Flip): Graphic =
+        graphic.copy(flip = newFlip)
+
+  given spatialGraphic(using bs: BasicSpatial[Graphic]): Spatial[Graphic] with
```
davesmith00000, line 120: This is good but I feel like you shouldn't need to implement all this stuff. You should be able to derive most of this generically for any `BasicSpatial[T]` no?

davesmith00000: The exception is the `transformTo` and `transformBy` methods, but maybe (🤔) either:
1) These are the only methods you have to hand roll for a `Spatial[T]` instance, or
2) They could be moved to `BasicSpatial`?

(2) doesn't feel unreasonable. They're not elegant methods designed to improve UX, they're there for efficiency.

indigo/indigo/src/main/scala/indigo/shared/scenegraph/Graphic.scala
bilki: It is building without errors so far :thinking:  running `build.sh` from the root folder (repo root to be more specific, not the indigo folder).

davesmith00000:
Friendly heads up: You'll need to rebase against master following release 0.8.1.

bilki:
It seems this experiment is going nowhere, I found no way to isolate the changes from the rest of the architecture, and I'm feeling like closing this draft. What do you think Dave? The typeclass rabbit hole is really deep in this case :smile:

davesmith00000:
I've been wondering about this myself. On reflection, I'm inclined to cut our loses. The problems are deep as you say and I'm not sure the effort will be worth the result in the end. How do you feel about it?

bilki:
Closing this, and maybe revisiting it in the future when (if) there is a chance to refactor this part of the architecture, although probably that will never happen, as it works perfectly fine as it is. It's just the hassle of pattern matching, which is unavoidable with the open class hierarchy.

davesmith00000:
Thanks @bilki. I would like to explore tightening it up again but I agree entirely with your summary - thanks for having a go at it! 👍

PR #164, opened 2021-06-27
Stop build artifacts from polluting git

AriFordsham:
I followed the build instructions and git was reporting 4K+ changes, all build artifacts. I added this and they all went away.


davesmith00000 (approved):
Many thanks! I can only assume you're using some workflow that uses coursier and I'm not, but hopefully that will help someone out. 🙂

Issue #167, opened 2021-07-02
Consider having Materials as a type parameter on Primitives.


davesmith00000:
The modifyMaterial method is a pain because you have to match the material type to know what modifications you can do.

Two ways to make that less painful:

1. A partial function - made total under the covers, not a big win in terms of ergonomics.

2. A type parameter for the material type, then modifyMaterial is effectively a map operation. Only draw back is a very slight more complex definition?

Give (2) a try and see how it feels?

PR #168, opened 2021-07-03
Built-in primitives have `Material` type param

davesmith00000:
This means that this:

```scala
      graphic
          .modifyMaterial {
            case m: LegacyEffects => m.withOverlay(Fill.Color(RGBA.Magenta.withAmount(0.75)))
            case m                => m
          }
```

...can now be this:

```scala
      graphic
          .modifyMaterial(_.withOverlay(Fill.Color(RGBA.Magenta.withAmount(0.75))))
```

At the cost of having to do this:

```scala
val foo: Graphic[_]
```

or this..

```scala
final case class (myGraphic: Graphic[Material.Bitmap])
```

...everywhere.

Worth it?

Applies to `Text`, `Graphic`, and `Sprite`. `TextBox`s and `Shape`s don't work with materials.


davesmith00000:
Fixes #167

Issue #161, opened 2021-06-24
More Texture Information in Shaders


davesmith00000:
It's currently really hard work to do anything out of the ordinary with textures. Simple example: You can't get the size of the texture, or the area it takes up on the texture atlas.

PR #170, opened 2021-07-03
Improved texture support in shaders

davesmith00000:
WIP

Fixes at least some of #161.

Need to review the roguelike starterkit to see if there's anything else I can do here.

Needs testing somehow...


davesmith00000:
Changes to shaders:

- TEXTURE_COORDS - is still available to avoid breaking things, but redundant, equal to UV.
- ATLAS_SIZE - is the size of the atlas on which your texture resides in pixels
- TEXTURE_SIZE - is the size of the texture in pixels
- FRAME_SIZE - (vertex only) is the size of the images's crop
- CHANNEL_0_POSITION - the position of the texture on the atlas in UV space
- CHANNEL_0_SIZE - the size of the texture on the atlas in UV space

PR #173, opened 2021-07-11
Add `#` prefix in RGBA.fromHexString

zetashift:
I just shadowed `hex`, not sure if this is the most idiomatic Scala.

`RGB.fromHexString` just calls the method of `RGBA` so this should also make it work without changing anything else.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
@@ -82,7 +82,6 @@ object RGBA:
     }
 
   def fromHexString(hex: String): RGBA =
-    val hex = if hex.startsWith("#") then hex.substring(1).trim else hex.trim
     hex match {
       case h if h.startsWith("0x") && h.length == 10 =>
         fromColorInts(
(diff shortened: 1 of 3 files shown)


indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
```diff
@@ -82,7 +82,8 @@ object RGBA:
     }
 
   def fromHexString(hex: String): RGBA =
-    hex.trim match {
+    val hex = if hex.startsWith("#") then hex.substring(1).trim else hex.trim
```
bilki, line 85: Shadowing produces a less readable code, imo. You could also use `tail` here instead of `substring(1)`, which could fail when the string is empty (not the case because it's nested within that `if`)

bilki (commented):
Looks useful, although I'm missing some test cases :test_tube: here:

https://github.com/PurpleKingdomGames/indigo/blob/5ebd0f897ab746218859458cc8793d294f0eb162/indigo/indigo/src/test/scala/indigo/shared/datatypes/RGBATests.scala#L24

indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
davesmith00000: There is also an established pattern here for this. Now I look at it I am wondering why on earth I didn't just use a regex, but nonetheless it would be good if all cases were extracted in the same manner.

davesmith00000:
The other thing is that we'll need to apply this to `RGB` - which will no doubt (but I haven't looked) have the same issue as `RGBA`.

zetashift:
Thank you for the feedback, I added the tests and made it more like the established pattern, let me know if anything is wrong!

davesmith00000 (approved):
That's looking much better - many thanks!

PR #178, opened 2021-07-16
Remove unreachable branch when traversing keyboard events

bilki:
Improved pattern matching exhaustivity in 3.0.1 is already paying off :smiley: Keyboard event is a sealed trait with two different events, this branch is unreachable.


davesmith00000 (approved):
Great! Thank you!

PR #198, opened 2021-09-29
Vertex, Vector and Point rotation

agdt3:
Added rotateBy for Vertex and Vector2
Added unit tests for both

Possible changes:
- Use Radians instead of Double
- Add rotateBy to Point as well


davesmith00000:
Looks great @agdt3! Regarding your two points:

1. Use Radians instead of Double

Yes definitely. Radians are an opaque type, so it costs nothing and provides more type safety. Plus that's how all the other APIs work so it reduces the surprise factor.

2. Add rotateBy to Point as well

Yes please.

davesmith00000 (approved):
Looks great - many thanks!

PR #232, opened 2021-10-31
#195 Add more built-in colours for both RGBA and RGB

### LINKED ISSUE #195: Add more built-in colours to RGBA & RGB
We have a bunch of pre-made colours like `RGBA.White`. I frequently want more simple colours when I'm prototyping, like orange and purple! What else is there?

sylvansson:
Close #195

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/datatypes/RGB.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/datatypes/RGB.scala
@@ -51,6 +51,24 @@ object RGB:
   val White: RGB   = RGB(1, 1, 1)
   val Black: RGB   = RGB(0, 0, 0)
 
+  val Coral: RGB     = RGBA.Coral.toRGB
+  val Crimson: RGB   = RGBA.Crimson.toRGB
+  val DarkBlue: RGB  = RGBA.DarkBlue.toRGB
+  val Indigo: RGB    = RGBA.Indigo.toRGB
+  val Olive: RGB     = RGBA.Olive.toRGB
+  val Orange: RGB    = RGBA.Orange.toRGB
+  val Pink: RGB      = RGBA.Pink.toRGB
+  val Plum: RGB      = RGBA.Plum.toRGB
+  val Purple: RGB    = RGBA.Purple.toRGB
+  val Salmon: RGB    = RGBA.Salmon.toRGB
+  val SeaGreen: RGB  = RGBA.SeaGreen.toRGB
+  val Silver: RGB    = RGBA.Silver.toRGB
+  val SlateGray: RGB = RGBA.SlateGray.toRGB
+  val SteelBlue: RGB = RGBA.SteelBlue.toRGB
+  val Teal: RGB      = RGBA.Teal.toRGB
+  val Thistle: RGB   = RGBA.Thistle.toRGB
+  val Tomato: RGB    = RGBA.Tomato.toRGB
+
   val Normal: RGB = White
   val None: RGB   = White
   val Zero: RGB   = RGB(0, 0, 0)
--- a/indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
@@ -68,6 +68,25 @@ object RGBA:
   val None: RGBA   = White
   val Zero: RGBA   = RGBA(0, 0, 0, 0)
 
+  // https://en.wikipedia.org/wiki/X11_color_names
+  val Coral: RGBA     = fromHexString("#FF7F50")
+  val Crimson: RGBA   = fromHexString("#DC143C")
+  val DarkBlue: RGBA  = fromHexString("#00008B")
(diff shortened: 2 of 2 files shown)


indigo/indigo/src/main/scala/indigo/shared/datatypes/RGB.scala
```diff
@@ -51,6 +51,24 @@ object RGB:
   val White: RGB   = RGB(1, 1, 1)
   val Black: RGB   = RGB(0, 0, 0)
 
+  val Coral: RGB     = RGBA.Coral.toRGB
```
sylvansson, line 54: @davesmith00000 I considered writing a function to convert instances of `RGBA` to `RGB` implicitly to make this section more concise but wasn't sure what casing would be best. `rgbaToRGB`?

indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
```diff
@@ -68,6 +68,25 @@ object RGBA:
...
+  val Plum: RGBA      = fromHexString("#DDA0DD")
+  val Purple: RGBA    = fromHexString("#A020F0")
+  val Salmon: RGBA    = fromHexString("#FA8072")
+  val SeaGreen: RGBA  = fromHexString("#2E8B57")
+  val Silver: RGBA    = fromHexString("#C0C0C0")
+  val SlateGray: RGBA = fromHexString("#708090")
+  val SteelBlue: RGBA = fromHexString("#4682B4")
+  val Teal: RGBA      = fromHexString("#008080")
+  val Thistle: RGBA   = fromHexString("#D8BFD8")
+  val Tomato: RGBA    = fromHexString("#FF6347")
```
sylvansson, line 88: @davesmith00000 Let me know if the colours I picked are a bit too "out-there" :-)

indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
```diff
@@ -68,6 +68,25 @@ object RGBA:
   val None: RGBA   = White
   val Zero: RGBA   = RGBA(0, 0, 0, 0)
 
+  // https://en.wikipedia.org/wiki/X11_color_names
+  val Coral: RGBA     = fromHexString("#FF7F50")
```
sylvansson, line 72: I figured using hex strings would be cleaner but I can rewrite this section to use doubles if needed.

davesmith00000 (approved):
This is interesting. It's not what I expected and yet it's also totally reasonable. 😁 

I had a quick google round at some of the colour tables, and you can get CSV files ready made with nearly a 1000 colours in them. Now, we _could_ suck in one of those CSV files at compile time and get sbt to just output a massive number of pre-baked colours into ready made class files - but! - is anyone _really_ going to know that colour "Vivid cerise" or "Office green" is without a preview?

So I think we can just stop there - give people a bit more choice and after that they're on their own. Also I think your choice of colours is great! I sincerely hope that everyone using auto-complete in their IDE chuckles: "Ha! Why they heck did they include `RGBA.Tomato`?!?". Perfect.

Many thanks. 👍

indigo/indigo/src/main/scala/indigo/shared/datatypes/RGB.scala
davesmith00000: Explicit is totally fine, preferable even.

indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
davesmith00000: Hex is fine thanks.

indigo/indigo/src/main/scala/indigo/shared/datatypes/RGBA.scala
davesmith00000: They're an exciting mix of out there while looking vaguely practical. 👍

sylvansson:
Just a few more colours and you could make yourself some [salmon in tomato-olive sauce](https://www.foodandwine.com/recipes/salmon-in-tomato-olive-sauce) ;-)

I assumed you simply wanted a handful of extra colours but I should have asked, sorry. I've never done any kind of code generation but I'd be more than happy to work on something more comprehensive if you'd like. If not, should I go ahead and merge or leave that to you?

davesmith00000:
All done - thanks again!

I'm not feeling the need for anything more comprehensive at the moment. Maybe in a few days after I've let it percolate for a while, but currently I'm very happy with the colour based recipe recommendation. ;-)

Code generation is good fun and occasionally super useful. I use it in cases where I'm always going to load in a big lump of data, parse it, and then have to search it for information. If you instead take that lump of data and just generate a module - which is just a text file - then you get to compile it right into your code base, full type safety, no loading.

Here's an example, if you're interested, it's a little thing that generates tile information for Dwarf Fortress style asset packs (which is just a grid of ascii symbols):

https://github.com/davesmith00000/roguelike-tutorial/blob/main/part1/project/TileCharGen.scala

Which is called from:

https://github.com/davesmith00000/roguelike-tutorial/blob/main/part1/build.sbt#L31

PR #260, opened 2021-12-17
Use `CloneTiles` for `Text` in WebGL 2

davesmith00000:
This PR is an attempt to make the standard `Text` primitive much less resource hungry.

The way the current version works, and will continue to work for WebGL 1.0, is that for each letter in the text to be rendered, and individual entity is spawned in the right location, with the right crop of the font sheet to draw that letter.

This has worked well for years now, with the caveat that it was never intended to work for large bodies of text. The advantage of `Text` over `TextBox` (which will support lots of text) is that `Text` is pixel perfect and `TextBox` cannot be because of the way browsers render text to canvas elements.

Unfortunately the game I'm working on needs lots of text. In the game, thanks to the [roguelike-starterkit](https://github.com/PurpleKingdomGames/roguelike-starterkit) we have 4 methods of drawing text, but the most convenient is still good old `Text` and so it needs to be faster. Faster here means "allocate less and therefore trigger fewer damaging Major GC pauses".

The solution is to replace the "one entity per letter" method with the use of the same mechanism used to render `CloneTiles`, which will at most result in one entity per _line_ of text, and possible one entity per text entity.

A secondary advantage to this is that we heavily cache the render pipeline transform results internally, and this approach results in fewer and smaller cache entries.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
@@ -52,7 +52,6 @@ import indigo.shared.shader.ShaderPrimitive
 import indigo.shared.shader.Uniform
 import indigo.shared.shader.UniformBlock
 import indigo.shared.time.GameTime
-import org.w3c.dom.css.Rect
 
 import scala.annotation.tailrec
 import scala.collection.immutable.HashMap
--- a/indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
@@ -64,13 +64,14 @@ final class DisplayObjectConversions(
     fontRegister: FontRegister
 ) {
 
-  implicit private val textureRefAndOffsetCache: QuickCache[TextureRefAndOffset] = QuickCache.empty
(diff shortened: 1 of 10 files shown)


davesmith00000:
I think this is good to go. I have it published locally and all seems to be working well.

indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
```diff
@@ -172,66 +186,93 @@ final class DisplayObjectConversions(
   def sceneNodeToDisplayObject(
       gameTime: GameTime,
       assetMapping: AssetMapping,
-      cloneBlankDisplayObjects: => HashMap[CloneId, DisplayObject]
-  )(sceneNode: SceneGraphNode): DisplayEntity =
+      cloneBlankDisplayObjects: => HashMap[CloneId, DisplayObject],
+      renderingTechnology: RenderingTechnology,
+      maxBatchSize: Int
+  )(sceneNode: SceneGraphNode): (DisplayEntity, scalajs.js.Array[(CloneId, DisplayObject)]) =
```
davesmith00000, line 192: This is the ugliest bit, the return type of `(DisplayEntity, scalajs.js.Array[(CloneId, DisplayObject)])`.
Technically, that js array could be an option since there's only ever one. However, option is a special case of list/array, and all we're going to do is smoosh them all together, never test for their presence, nor do we need a null check. So js array should be the fastest collection type in this case.

PR #282, opened 2022-01-10
Add topLeft utility function to Camera.scala.

Sintrastes:
This patch adds a method to `Camera` that calculates the top left pixel of the camera when given the current viewport.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/scenegraph/Camera.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/scenegraph/Camera.scala
@@ -2,11 +2,13 @@ package indigo.shared.scenegraph
 
 import indigo.shared.datatypes.Point
 import indigo.shared.datatypes.Radians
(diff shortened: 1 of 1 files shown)


indigo/indigo/src/main/scala/indigo/shared/scenegraph/Camera.scala
```diff
@@ -65,6 +68,8 @@ object Camera:
   final case class LookAt(target: Point, zoom: Zoom, rotation: Radians) extends Camera:
     val isLookAt: Boolean = true
     val position: Point   = target
+    def topLeft(viewport: GameViewport) = 
+      target - Point(viewport.width/2, viewport.height/2) / zoom.toDouble.toInt
```
davesmith00000, line 72: Don't worry about it this time, but you should install Scalafmt, it'll keep all the code consistently formated.

davesmith00000 (approved):
Looks good - thanks! 😄

PR #284, opened 2022-01-12
Add mouse multi-button support to mouse event

bilki:
My initial idea was to only add support for the right mouse button, but it looks like it can be generic enough to support more than only left and right buttons.

This PR is still in progress, as I need to add tests, refactor the `Mouse` class, as it's now too much left-button-biased, and probably bake in any other ideas from the review feedback.

### DIFF
--- a/examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
+++ b/examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
@@ -34,8 +34,8 @@ object Confetti extends IndigoDemo[Unit, Unit, Model, Unit]:
     Outcome(())
 
   def updateModel(context: FrameContext[Unit], model: Model): GlobalEvent => Outcome[Model] = event =>
-    val isLeftDown  = context.mouse.isButtonDown(MouseButton.LeftMouseButton)
-    val isRightDown = context.mouse.isButtonDown(MouseButton.RightMouseButton)
+    val isLeftDown  = context.mouse.isLeftDown
+    val isRightDown = context.mouse.isRightDown
 
     event match
       case FrameTick if isLeftDown || isRightDown =>
--- a/examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
+++ b/examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
@@ -1,7 +1,6 @@
 package com.example.confetti
 
 import indigo._
-import indigo.shared.events.MouseButton
 import indigoextras.subsystems.FPSCounter
 
 import scala.scalajs.js.annotation.JSExportTopLevel
@@ -34,24 +33,27 @@ object Confetti extends IndigoDemo[Unit, Unit, Model, Unit]:
   def initialViewModel(startupData: Unit, model: Model): Outcome[Unit] =
     Outcome(())
 
-  private def maybeMouseUp(mouse: Mouse): Option[LmbOrRmb] =
-    if mouse.released(MouseButton.LeftMouseButton) then Some(MouseButton.LeftMouseButton)
-    else if mouse.released(MouseButton.RightMouseButton) then Some(MouseButton.RightMouseButton)
-    else None
+  def updateModel(context: FrameContext[Unit], model: Model): GlobalEvent => Outcome[Model] = event =>
+    val isLeftDown  = context.mouse.isButtonDown(MouseButton.LeftMouseButton)
+    val isRightDown = context.mouse.isButtonDown(MouseButton.RightMouseButton)
 
-  def updateModel(context: FrameContext[Unit], model: Model): GlobalEvent => Outcome[Model] =
-    case FrameTick =>
-      maybeMouseUp(context.mouse) match
-        case Some(source) =>
-          Outcome(
-            model
-              .spawn(context, 15, source)
-              .update
-          )
-        case None =>
-          Outcome(model.update)
-    case _ =>
-      Outcome(model)
+    event match
+      case FrameTick if isLeftDown || isRightDown =>
+        (isLeftDown, isRightDown) match
+          case (true, true) =>
+            val lmb = model.spawn(context, 8, MouseButton.LeftMouseButton)
+            val rmb = lmb.spawn(context, 7, MouseButton.RightMouseButton)
+            Outcome(rmb.update)
+          case (true, false) =>
+            Outcome(model.spawn(context, 15, MouseButton.LeftMouseButton).update)
+          case (false, true) =>
+            Outcome(model.spawn(context, 15, MouseButton.RightMouseButton).update)
+          case (false, false) => // Not reachable
+            Outcome(model.update)
+      case FrameTick =>
+        Outcome(model.update)
+      case _ =>
+        Outcome(model)
 
   def updateViewModel(context: FrameContext[Unit], model: Model, viewModel: Unit): GlobalEvent => Outcome[Unit] =
     _ => Outcome(viewModel)
@@ -73,9 +75,8 @@ object Confetti extends IndigoDemo[Unit, Unit, Model, Unit]:
       .withFontSize(Pixels(12))
       .withColor(RGBA.White)
 
-  def helpText(maybeButton: Option[MouseButton]): TextBox =
-    val msg = maybeButton.fold("Click anywhere!")(btn => s"$btn clicked!")
-    TextBox(msg, 640, 20).alignRight
+  val helpText: TextBox =
+    TextBox("Click left, right or both anywhere!", 640, 20).alignRight
       .withFontSize(Pixels(12))
       .withColor(RGBA.White)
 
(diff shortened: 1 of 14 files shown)


bilki:
This was finally a bit more involved than I thought, as the canvas `onclick` only listens to the click event from the left mouse button, so I needed to steer midway through the implementation.

The implementation basically maintains all the existing functions for LMB (to avoid breaking any existing games and examples), while adding a few more generic functions accepting a `MouseButton` as input. Still missing some tests that probably will fall under `InputStateTests`.

I also modified the `confetti` example in order to manually test that everything was working as intended. I had to split the color dots array in half, assigning red/green to LMB, and blue/yellow to RMB.

Any feedback will be greatly appreciated and welcomed.

examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
```diff
@@ -1,6 +1,7 @@
 package com.example.confetti
 
 import indigo._
+import indigo.shared.events.MouseButton
```
davesmith00000, line 4: If you haven't already, could you please add an alias to this inside `package.scala`? Then you won't need this import.

examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
```diff
@@ -33,27 +34,36 @@ object Confetti extends IndigoDemo[Unit, Unit, Model, Unit]:
...
-    case FrameTick if context.mouse.leftMouseIsDown =>
-      Outcome(
-        model
-          .spawn(context, 15)
-          .update
-      )
+  private def maybeMouseUp(mouse: Mouse): Option[LmbOrRmb] =
+    if mouse.released(MouseButton.LeftMouseButton) then Some(MouseButton.LeftMouseButton)
+    else if mouse.released(MouseButton.RightMouseButton) then Some(MouseButton.RightMouseButton)
+    else None
```
davesmith00000, line 40: Feels like this could be a general function on `Mouse`? (I'm reading top to bottom so apologies if this exists further down).

examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
```diff
@@ -63,32 +73,39 @@ object Confetti extends IndigoDemo[Unit, Unit, Model, Unit]:
       .withFontSize(Pixels(12))
       .withColor(RGBA.White)
 
-  val helpText: TextBox =
-    TextBox("Click anywhere!", 640, 20).alignRight
+  def helpText(maybeButton: Option[MouseButton]): TextBox =
+    val msg = maybeButton.fold("Click anywhere!")(btn => s"$btn clicked!")
```
davesmith00000, line 77: "Click left and right mouse anywhere!" ?

indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
```diff
@@ -53,7 +55,8 @@ object WorldEvents {
       globalEventStream.pushGlobalEvent(
         MouseEvent.MouseDown(
           absoluteCoordsX(e.clientX) / magnification,
-          absoluteCoordsY(e.clientY) / magnification
+          absoluteCoordsY(e.clientY) / magnification,
+          MouseButton.fromOrdinal(e.button)
```
davesmith00000, line 59: If you had a fancy mouse with more than 5 buttons, this would result in a `NoSuchElementException`. Could we add a similar function to the `MouseButton` companion that checks the range is valid first? Ideally, I feel like hitting a funny button would do nothing (or act like a left click?) rather than blow up. What do you think?

indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
```diff
@@ -62,11 +65,15 @@ object WorldEvents {
...
         )
       )
     }
 
+    // Prevent right mouse button from popping up the context menu
+    canvas.oncontextmenu = _.preventDefault()
```
davesmith00000, line 75: I think that's fair in general. The alternative would be for this to be the default, but to add a flag in the `AdvancedConfig` so the game dev can switch it on and off. I don't feel strongly about it though - you?

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -87,14 +87,24 @@ case object FullScreenExited extends ViewEvent
...
+
 /** Represents all mouse events
   */
 sealed trait MouseEvent extends InputEvent {
   val position: Point
   val x: Int
   val y: Int
 }
-object MouseEvent {
+object MouseEvent:
```
davesmith00000, line 107: Thanks - spotted that the other day but didn't fix it! 😄

indigo/indigo/src/main/scala/indigo/shared/input/Mouse.scala
```diff
@@ -91,17 +121,13 @@ object Mouse {
...
+      case MouseEvent.MouseDown(_, MouseButton.LeftMouseButton) :: xs =>
         isLeftMouseDown(true, xs)
-
-      case MouseEvent.MouseUp(_) :: xs =>
+      case MouseEvent.MouseUp(_, MouseButton.LeftMouseButton) :: xs =>
         isLeftMouseDown(false, xs)
-
       case _ :: xs =>
         isLeftMouseDown(isDown, xs)
-    }
```
davesmith00000, line 106: Is this whole function (which I've never liked) now made redundant by some other function? I've got a feeling it isn't? But:

1. If it is redundant, could we add a deprecation annotation telling the user which function to use instead?
2. If it isn't... do we need one for right click? ...and the others? This does get used all over the place unfortunately.

Also if it's staying, suggestions for a better name welcome!

davesmith00000: This is used to work out if the left mouse has been continuously held down, in case that isn't clear.

davesmith00000 (commented):
Looks really good! I've left a couple of comments and questions.

Before I fully approve and merge it though, I feel like I'd be remiss for not doing a bit of QA testing. Will try and do that soon.

Thanks! 👍

examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
davesmith00000: Your `(btn => s"$btn clicked!")` is bearly perceptible because it only flashes up for a moment.

examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
```diff
@@ -33,27 +34,36 @@ object Confetti extends IndigoDemo[Unit, Unit, Model, Unit]:
   def initialViewModel(startupData: Unit, model: Model): Outcome[Unit] =
     Outcome(())
 
-  def updateModel(context: FrameContext[Unit], model: Model): GlobalEvent => Outcome[Model] =
-    case FrameTick if context.mouse.leftMouseIsDown =>
```
davesmith00000, line 37: So the removal of this bit breaks the demo. It shows that your code works (red+green for left, blue+yellow for right), but the demo now only allows for clicks, not hold downs. Can we allow hold down of either button?

davesmith00000:
I like the left / right aspect of the confetti demo, but I'd like to bring back hold down actions. Feels a bit broken currently.

examples/confetti/src/main/scala/com/example/confetti/Confetti.scala
bilki: Hmmm, feels like this could be added to `Mouse` as something like `def maybeMouseReleased(mouse: Mouse): Option[MouseButton]`. The problem with that function is that it returns the "first" button released (or pressed) in the events list. Not sure how useful could that be, even the behavior could be a bit surprising for the user because the order or the `if`s inside that function would be arbitrary 🤔

indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
bilki: Can we do this in another PR? 🤔 I think it makes sense to let the user decide about it, perhaps there are use cases I can not think of right now, i.e. something like integrating a game with the rest of a web page... I'm totally making up the use case though.

indigo/indigo/src/main/scala/indigo/shared/input/Mouse.scala
bilki: Revamped this function into a more generic one, in a similar fashion to what's done for the `Keyboard`... I'll push it and let you review that new impl 👀

bilki:
Fixed the confetti demo, I'm afraid of that demo being too much convoluted with my changes 😓 Now the confetti spawns when clicking left, right, or both (combining dots in that case).

Let me know what you think of the changes I made to `Mouse`. The bit I'm not very sure about is deprecating `isLeftMouseDown`, but it seems like the right move, now that the class supports more buttons... Added `nowarn` to the UI components, those changes need to be done in a future PR, I think the same one that will remove `isLeftMouseDown` could be a good candidate.

indigo/indigo-extras/src/main/scala/indigoextras/ui/Button.scala
```diff
@@ -15,6 +15,9 @@ import indigo.shared.scenegraph.Sprite
 import indigo.shared.scenegraph.Text
 import indigo.shared.scenegraph.TextBox
 
+import scala.annotation.nowarn
+
+@nowarn("msg=value leftMouseIsDown in class Mouse is deprecated")
```
davesmith00000, line 20: Oh I've been using `@deprecated` above the method?

davesmith00000 (approved):
Looks great. I've tested it and it appears to work perfectly - great job! Thanks 😄

PR #307, opened 2022-01-23
Remove Model/View update skipping

davesmith00000:
~~**Needs further testing.**~~ (done)

I've been considering removing the option to disable update and view skipping for a while, largely because it seems unnecessary.

Removing it does mean that Indigo makes no attempt to smooth out the frames, and leaves it to the browser to decide when to drop frames ...but the browser does that anyway.

This code was also masking a nasty bug.

Say your game is running at 60 fps, this means you have 16.7ms to do everything. Prior to this PR, if Indigo notices the time delta between the last frame and this one was greater than (16.7 * 2), then it would skip this update - but not before draining the event queue, resulting in skipped events. This was really only noticeable on games with really regular/long frame updates, typically exacerbated by the browser dropping frames at the same time. This all combined made it noticeable that events were being lost somewhere, but the circumstances were very hard to nail down.

Part of the reason this bug has crept in is because the skipping mechanism has been in place since the beginning of time, but game loop process changed a lot when the notion of frame processors arrived.

This PR removes the ability to skip updates, simplifies the code, and removes the need for the frame processor to have any notion of running without rendering.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
+++ b/indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
@@ -47,101 +47,71 @@ class GameLoop[StartUpData, GameModel, ViewModel](
       val startUpData: StartUpData =
         gameEngine.startUpData
 
-      // Model updates cut off
-      if (gameConfig.advanced.disableSkipModelUpdates || timeDelta < gameConfig.haltModelUpdatesAt) {
+      val gameTime =
+        new GameTime(Millis(time).toSeconds, Seconds(timeDelta.toDouble / 1000d), gameConfig.frameRate)
+      val collectedEvents = gameEngine.globalEventStream.collect ++ List(FrameTick)
+      val dice            = Dice.fromSeconds(gameTime.running)
 
-        val gameTime =
-          new GameTime(Millis(time).toSeconds, Seconds(timeDelta.toDouble / 1000d), gameConfig.frameRate)
-        val collectedEvents = gameEngine.globalEventStream.collect ++ List(FrameTick)
-        val dice            = Dice.fromSeconds(gameTime.running)
+      // Persist input state
+      inputState = InputState.calculateNext(
+        inputState,
+        collectedEvents.collect { case e: InputEvent => e },
+        gameEngine.gamepadInputCapture.giveGamepadState
+      )
 
-        // Persist input state
-        inputState = InputState.calculateNext(
+      val processedFrame: Outcome[(GameModel, ViewModel, SceneUpdateFragment)] =
+        frameProcessor.run(
+          startUpData,
+          gameModelState,
+          viewModelState,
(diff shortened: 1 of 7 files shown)


indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
```diff
@@ -47,101 +47,71 @@ class GameLoop[StartUpData, GameModel, ViewModel](
...
-              throw e
-
-            case Outcome.Result(state, globalEvents) =>
-              gameModelState = state._1
-              viewModelState = state._2
-              globalEvents.foreach(e => gameEngine.globalEventStream.pushGlobalEvent(e))
-          }
+      // Persist frame state
+      val scene =
+        processedFrame match {
```
bilki, line 76:
```suggestion
        processedFrame match
```

indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
```diff
@@ -47,101 +47,71 @@ class GameLoop[StartUpData, GameModel, ViewModel](
...
+            IndigoLogger.error("The game has crashed...")
+            IndigoLogger.error(oe.reportCrash)
+            throw e
+
+          case Outcome.Result(state, globalEvents) =>
+            gameModelState = state._1
+            viewModelState = state._2
+            globalEvents.foreach(e => gameEngine.globalEventStream.pushGlobalEvent(e))
+            state._3
         }
```
bilki, line 87:
```suggestion
          case Outcome.Result((gameModel, viewModel, sceneUpdateFragment), globalEvents) =>
            gameModelState = gameModel
            viewModelState = viewModel
            globalEvents.foreach(e => gameEngine.globalEventStream.pushGlobalEvent(e))
            sceneUpdateFragment
```

indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
```diff
@@ -47,101 +47,71 @@ class GameLoop[StartUpData, GameModel, ViewModel](
...
+        gameEngine.renderer.renderingTechnology,
+        gameConfig.advanced.batchSize
+      )
+
+      // Render scene
+      gameEngine.renderer.drawScene(sceneData, gameTime.running)
+
+      // Tick
       gameEngine.platform.tick(gameEngine.gameLoop(time))
     } else gameEngine.platform.tick(loop(lastUpdateTime))
```
bilki, line 116: Just a side note, it's a bit hard to follow the `if` for this `else` branch. I'd consider moving the whole `if` branch to a function 🤔

bilki (commented):
Only a few cosmetic comments 👀

davesmith00000:
Thanks for the nudge @bilki. I've taken this opportunity to do some related tidying up.

Issue #290, opened 2022-01-16
Add support for mouse scroll bar events


Sintrastes:
A lot of games make use of this sort of input to event, so this is important to have. Some examples:

   * Zooming in and out
   * Scrolling up and down in a UI for a settings page
   * Toggling through the currently selected item in an item bar.

Some links from @bilki for some relevant docs: https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent, https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event

davesmith00000:
I believe @bilki is looking into this? But let me know if that's not the case! 😄

Sintrastes:
That would be good to know! I was under the impression he was just working on mouse click events, but I could be wrong. Don't want to duplicate work.

bilki:
Yes, I'm working only the button click actions, as it was relatively easy to extend the current set of events. Adding the mouse wheel/scroll actions is a bit different, I'm not sure how much effort would it take, and it feels like #284 is already a self-contained change, we could tackle this other feature later in another PR.

davesmith00000:
Apologies - didn't mean to give you more work there! 😅

bilki:
I'm going to give this a try, give me a shout if you already implemented something 👀

Sintrastes:
@bilki Nothing here yet. Go for it!

bilki:
Solved in #308

PR #308, opened 2022-01-23
Add mouse wheel events for scrolling up and down

bilki:
Hopefully, this will solve #290. I'm currently experimenting with a very simple interface exposed to the dev using a `scrolled` function in `Mouse` that returns an `Option` that contains the scrolling direction if any wheel event exists. I think this is the most basic strategy, as providing the user with more detail implies knowing what the delta mode (pixels, lines, etc.) is, which is implementation-specific to browser and device... What do you think, is it worth taking more time to precisely determine the pixels, in case some games require a finer handling of wheel/scrolling events?

I added a new example to the `examples` folder for the sole purpose of testing and showing how mouse events are handled.


indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -91,11 +91,12 @@ case object FullScreenExitError extends ViewEvent
...
-  case MiddleMouseButton    extends MouseButton
-  case RightMouseButton     extends MouseButton
-  case BrowserBackButton    extends MouseButton
-  case BrowserForwardButton extends MouseButton
+  case LeftMouseButton, MiddleMouseButton, RightMouseButton, BrowserBackButton, BrowserForwardButton
+
+/** Represents in which direction the mouse wheel was rotated
+ */
+enum MouseWheelDir derives CanEqual:
+  case Up, Down
```
davesmith00000, line 99: In general, I avoid abbreviations. Here I would probably prefer either `MouseWheel.Up` or maybe better(?) `MouseWheel.ScrollUp`.

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -181,6 +182,21 @@ object MouseEvent:
     def apply(x: Int, y: Int): Move =
       Move(Point(x, y))
 
+  /** The mouse wheel was rotated a certain amount into the Y axis.
+    *
+    * @param x
+    *   X coord relative to magnification level
+    * @param y
+    *   Y coord relative to magnification level
```
davesmith00000, line 190: These are wrong, I imagine they're a copy paste and the real person at fault is me from when I changed the `Click` event to hold a `Point`. 😄 

Also needs the amount field,

davesmith00000 (commented):
This is looking really good! I've left two small comments.

indigo/indigo/src/main/scala/indigo/shared/input/Mouse.scala
```diff
@@ -47,6 +48,20 @@ final class Mouse(
...
+  }
+
+  lazy val scrolled: Option[MouseWheelDir] =
+    if wasScrolled then
+      val amount = mouseEvents.foldLeft(0d) {
+        case (acc, MouseEvent.Wheel(_, deltaX)) => acc + deltaX
+        case (acc, _)                           => acc
+      }
+      if amount < 0 then Some(MouseWheelDir.Up) else Some(MouseWheelDir.Down)
+    else Option.empty[MouseWheelDir]
```
davesmith00000, line 63: Probably fine because it's probably only a handful of events and maybe 2 wheel events per frame, so this is just for future reference: A general performance point in the engine is to avoid `foldLeft` on `List` as I've found it to be very slow.

davesmith00000: I think the first check is unnecessary, and possibly amounts to iterating over the list anyway?

```scala
  lazy val scrolled: Option[MouseWheelDir] =
    val amount = mouseEvents.foldLeft(0d) {
      case (acc, MouseEvent.Wheel(_, deltaX)) => acc + deltaX
      case (acc, _)                           => acc
    }
    if amount == 0 then None
    else if amount < 0 then Some(MouseWheelDir.Up)
    else Some(MouseWheelDir.Down)
```

I don't feel strongly about it, just trying it out to see how it feels.

davesmith00000:
Regarding your question about the user interface, I think this is a good starting point.

Best case someone finds it useful, worst case someone raises and issue asking for something more sophisticated, and we have a good starting point to add the required functionality.

If you are happy with it, I'd keep it as is and wait for user feedback. Having said that, I haven't tried the new example yet so the first user might be me. 😄

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
bilki: Copy-pasted 😅

bilki:
Oh, I forgot to review if we need to add any docs for this 👀

indigo/docs/gameloop/events.md
```diff
@@ -57,6 +57,8 @@ You can create your own events by simply extending `GlobalEvent`.
 
 ### `InputEvent`s
 
+Handling `InputEvent`s can be a bit tricky in some situations, so Indigo includes `Mouse` and `Keyboard` classes that can be accessed from the [frame context](gameloop/frame-context.md), providing a rich interface to gather more complex information from those input devices.
```
davesmith00000, line 60: 👍

davesmith00000 (approved):
Looks great and the demo works perfectly!

PR #309, opened 2022-01-24
Supply sbt & Mill versions to MDoc

davesmith00000:
Land after https://github.com/PurpleKingdomGames/indigo/pull/305


bilki (dismissed):
👍

PR #310, opened 2022-01-24
Upgrade to Mill 0.10.0

davesmith00000:
Relates to issue #292.

Land PR https://github.com/PurpleKingdomGames/indigo/pull/305 first. (Hence the draft PR)

This is a surprisingly breaking change that even affects the sbt plugin.

Mill 0.10.0 brings in some superficial changes, the one that affects us is a change of structure to the `out` directory from `(..)/fastOpt/dest/(..)` to `(..)/fastOpt.dest/(..)`.

That's ok, I briefly added a crude condition to pick the right one depending on the Mill version.

However, Mill is part of the Li Haoyi ecosystem (just like any other ecosystem) and so they've also upgraded `os-lib` to version `0.8.0` (previously we used `0.7.8` everywhere). So if you're on version `0.10.0` you get one os-lib, and if you're on a previous version you get a different `os-lib`.. and they are not binary compatible.

So after the next release, everyone will need to upgrade to Mill `0.10.0`.


davesmith00000:
Marked as ready for review, but I'm just going to rebase against master and try it with the new build work.

PR #313, opened 2022-01-26
Fixed #128: Add standard modifiers to UI elements

### LINKED ISSUE #128: Moving buttons is bothersome
Fallout of reorganising the `SceneNode` hierarchy: `ButtonAssets` are now a `SceneNode` and have no standard methods to reposition them. You have to match on the sub-type and move that. 😒

davesmith00000:
I've added some of the standard modifiers you find in Indigo. The button move is a bit deceptive though, you have to move the button in the model where it's stored. If you move it just before it's rendered, then the hit area doesn't move with it. Still it's an improvement.

### DIFF
--- a/indigo/indigo-extras/src/main/scala/indigoextras/ui/Button.scala
+++ b/indigo/indigo-extras/src/main/scala/indigoextras/ui/Button.scala
@@ -4,6 +4,7 @@ import indigo.shared.Outcome
 import indigo.shared.datatypes.Depth
 import indigo.shared.datatypes.Point
 import indigo.shared.datatypes.Rectangle
+import indigo.shared.datatypes.Size
 import indigo.shared.events.GlobalEvent
 import indigo.shared.input.Mouse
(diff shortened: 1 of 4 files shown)


bilki:
> The button move is a bit deceptive though, you have to move the button in the model where it's stored.

A bit counterintuitive, yes. Do you mean in the `updateModel` function and/or the `updateViewModel` function?

davesmith00000:
Yeah so, think about the order of events:

Update model -> detect mouse click inside button at (10, 10)
Update view model -> ...
Present -> move the button to (50, 50)

Uh oh, I detected the mouse click in one location and then drew it in another.

The solution is to move it in the place it's stored (i.e. one of the models). Which is weird. But the UI components are odd conceptually.

I think there's an argument that we need a new construct entirely for UI elements specifically, this PR isn't fixing the underlying problems, just the specifics of `Button`.

bilki (dismissed):
✅ looks good.

Issue #316, opened 2022-01-29
Mouse position should be relative to the canvas


hobnob:
At the moment the mouse position is relative to the page, which gives weird results when the canvas is not at 0,0 on the page. We should deal with this by giving coordinates relative to the canvas instead. This StackOverflow details how we can do this, and also deals with what happens when the canvas is scaled etc:
https://stackoverflow.com/questions/17130395/real-mouse-position-in-canvas

I'm less fussed about the scaling and rotation. Certainly for now it will be enough to deal with the offset.

PR #317, opened 2022-01-29
More accurate mouse positions

hobnob:
The mouse position was relative to the page rather than relative to the canvas (see issue #316). This fixes that for unscaled and untransformed canvas elements such that the mouse position is now relative to the canvas in all mouse events.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
+++ b/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
@@ -35,10 +35,12 @@ object WorldEvents {
   def init(canvas: html.Canvas, magnification: Int, globalEventStream: GlobalEventStream): Unit = {
     // Onclick only supports the left mouse button
     canvas.onclick = { (e: dom.MouseEvent) =>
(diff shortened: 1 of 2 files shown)


davesmith00000:
Thanks @hobnob, looks good!

Couple of things:

1. Looks like you need to rebase the branch against master and re-push it, as there are merge conflicts.
2. Testing: I had a quick look at MDN (https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) but their demos don't make a lot of sense at face value. Is testing this as simple as changing the page margin/padding or embedding it under some element and checking the mouse coords look ok?

bilki:
Nothing to add to what @davesmith00000 mentioned, but just a little reminder that some other mouse events have been added after the commit parent for this PR so, once you rebase onto master, those need an update as well for their coords.

hobnob:
All rebased and the events updated :-)
@davesmith00000 Yup, testing is just a case of having a scrollable body by introducing a padding, margin, or large element

davesmith00000 (approved):
LGTM - thanks @hobnob 😄

PR #318, opened 2022-01-31
Adds a blog entry for updating IndigoSandbox to IndigoGame

hobnob:

PR #320, opened 2022-02-01
Hello Indigo + Tyrian guide

hobnob:
This relies on #318 as it links directly to it, and also relies on the Tyrian + Indigo subsystem. I fully expect the interface for this subsystem to change, but this PR is here for general wording approval and will be updated once the API has been agreed on and published.

It should not be merged until the sub-system and #318 have both been published

### DIFF
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -24,6 +24,23 @@ In this guide we'll take the basic `hello-indigo` example and add a button and
 a counter that scales with the size of the screen. It's a pretty basic example,
 but it should give you an idea of what can be achieved.
 
+## Overview
+
+This is a fairly involved guide with many parts to it that need to be implemented
+in order to get everything working. However the concept is pretty simple - we'll
+be using the Tyrian Bridge as a communication layer between Indigo and our
+`index.html` (which is controlled by Tyrian).
+
+These are the steps we're going to take:
+
+* [Set up the environment](#setting-up-the-environment) - e.g. clone `hello-indigo` and set up ParcelJS
+* [Modify our build files](#setting-up-the-build) - e.g. update our `build.sbt` or `build.sc` files
+* [Create our static files](#setting-up-the-html-and-parceljs) - e.ge create our html, js, css and other static content
+* [Update Indigo with a new subsystem](#update-helloindigoscala) and [Initialise our Tyrian app](#create-a-tyrian-app)
+* [Get Tyrian publishing messages to Indigo](#tyrian-to-indigo-communication) (and Indigo to subscribe to them)
+* [Get Indigo publishing messages to Tyrian](#indigo-to-tyrian-communication) (and Tyrian to subscribe to them)
+* And finally [creating a responsive UI in CSS](#responsive-ui)
+
 ## Setup
 
 ### Setting up the Environment
@@ -206,7 +223,7 @@ adding an entry to the `boot` method like so:
      )
 ```
 
-### Create a Tyrian App
+## Create a Tyrian App
 
 We'll now develop our initial Tyrian app, which will consist simply of
 a canvas (for Indigo), a counter, and a button, wrapped in a few `div` elements.
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -126,6 +126,10 @@ We'll also need an `index.html` to hold our basic HTML data:
 </html>
 ```
 
+Tyrian currently uses `snabbdom` for adding and removing elements in a web page.
+To do this we will need to add a dependency to `snabbdom` by running
+`yarn add snabbdom`.
+
 You'll notice we're using the direct JS output from our build here, which may
 feel odd. What will happen when we run ParcelJS through Yarn is that the HTML
 will be copied to a build directory along with the JS, CSS and any dependant
@@ -155,10 +159,8 @@ following to `package.json`:
 }
 ```
 
-We need to add a couple of dependencies. Firstly we'll add the
-dependency for static files by running
-`yarn add parcel-reporter-static-files-copy --dev` from the command line; and
-secondly we'll add `snabbdom` for Tyrian by running `yarn add snabbdom`.
+We need to add the dependency for static files by running
+`yarn add parcel-reporter-static-files-copy --dev` from the command line.
 
 Finally, we need to configure ParcelJS so that it knows how to copy our static
 files. Add a `.parcelrc` file and add the following:
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -204,7 +204,7 @@ like this:
 ```diff
      msg match
 +      case Msg.AddDot =>
-+        (model, model.bridge.publish(IndigoGameId(gameDivId), ""))
++        (model, model.bridge.publish(IndigoGameId(gameDivId), 0))
        case Msg.StartIndigo =>
 ```
 
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -20,8 +20,8 @@ Using Tyrian we can utilise HTML and CSS to ensure our overlay UI scales
 no matter what screen size we use.
 
 In this guide we'll take the basic `hello-indigo` example and add a button and
-a counter that scales with the size of the screen. It's a pretty basic example to
-be honest, but it should give you an idea of what can be achieved.
+a counter that scales with the size of the screen. It's a pretty basic example,
+but it should give you an idea of what can be achieved.
 
 ## Setup
 
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -98,12 +98,38 @@ Usually Indigo will generate all of our HTML for us and run it through Electron,
 but for this example we're going to generate our own HTML so that we can
 inject CSS.
 
-Firstly we'll create an `index.html` and `app.js` file that looks like
-[this](https://gist.github.com/hobnob/eaa03bfe2da14562e5f2819078f16d63). You'll
-notice we're using the direct JS output from our build here, which may feel odd.
-What will happen when we run ParcelJS through Yarn is that the HTML will be
-copied to a build directory along with the JS, CSS and any dependant static
-files (such as images) that may be needed.
+Firstly we'll create an `app.js` file that simply loads Tyrian and then launches
+it for our page:
+
+```js
+import {
+    TyrianApp
+} from './out/HelloIndigo/fastOpt.dest/out.js';
+
+TyrianApp.launch("main");
+```
+
+We'll also need an `index.html` to hold our basic HTML data:
+
+```html
+<!DOCTYPE html>
+<html>
+  <head>
+      <meta charset="UTF-8">
+      <title>Hello Indigo</title>
+      <link rel="stylesheet" href="css/main.css" />
+      <script type="module" defer src="app.js"></script>
+  </head>
+  <body>
+      <div id="main"></div>
+  </body>
+</html>
+```
+
+You'll notice we're using the direct JS output from our build here, which may
+feel odd. What will happen when we run ParcelJS through Yarn is that the HTML
+will be copied to a build directory along with the JS, CSS and any dependant
+static files (such as images) that may be needed.
 
 Now generate an empty `css/main.css`.
 
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -13,8 +13,9 @@ These tend to scale with the size of the screen and the position of such
 elements may change depending on the capabilities of the display. Many game
 engines provide a way to create overlay UI elements that scale in some way to
 the space they're in using a separate UI system. In Indigo we're exporting to
-JavaScript so it makes sense for us to use other web technologies for a problem
-that is already, ultimately, solved.
+JavaScript and a browser (even in Electron), so depending on how we want the UI
+to look, one option is to employ more standard web technologies to do this work
+for us, i.e. HTML and CSS.
 
 Using Tyrian we can utilise HTML and CSS to ensure our overlay UI scales
 no matter what screen size we use.
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -35,7 +35,7 @@ that then rotates around the canvas. We'll modify this so that clicking a button
 does this job instead.
 
 You'll want to upgrade `hello-indigo` to use `IndigoGame`. To do this either
-follow [this guide]()
+follow [this guide](howto-indigo-game.md)
 or replace `HelloIndigo.scala` with
 [this](https://gist.github.com/hobnob/c24f00936e91a7b7e5d644d19e4f1b32)
 
--- a/indigo/docs/guides/howto-hello-indigo-tyrian.md
+++ b/indigo/docs/guides/howto-hello-indigo-tyrian.md
@@ -68,8 +68,8 @@ Next you'll need to update either your
 -    }
 -  }
 -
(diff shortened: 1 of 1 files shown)


indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+
+For this guide we'll be using the `hello-indigo` example in either
+[mill](https://github.com/PurpleKingdomGames/hello-indigo) or
+[sbt](https://github.com/PurpleKingdomGames/hello-indigo-sbt).
+In the existing implementation clicking on the dots will create a new dot
+that then rotates around the canvas. We'll modify this so that clicking a button
+does this job instead.
+
+You'll want to upgrade `hello-indigo` to use `IndigoGame`. To do this either
+follow [this guide]()
```
hobnob, line 38: @davesmith00000 Need to add a link to the relevant guide here

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
++     ivy"io.indigoengine::tyrian-indigo-bridge::$tyrianVersion",
+      ivy"io.indigoengine::indigo-json-circe::$indigoVersion",
+      ivy"io.indigoengine::indigo::$indigoVersion",
+      ivy"io.indigoengine::indigo-extras::$indigoVersion"
+    )
+```
+
+ or `build.sbt` (if using sbt) to this:
+
+ ```diff
```
hobnob, line 86: Need to do a relevant diff for SBT - Not sure I'm confident enough to do one @davesmith00000

davesmith00000 (approved):
Great job @hobnob!

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
+---
+id: howto-responsive-ui-with-tyrian
+title: How to make a custom entity
```
davesmith00000, line 3: Needs updating

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+There tend to be two primary types of UI in games: In-game UI (such as health over a
+player), and overlay (such as a HUD or dialog box). In-game UI tends not to
+worry too much about the screen size, and is dealt with primarily by the
+magnification configuration within Indigo. But what about overlays and HUD's?
+These tend to scale with the size of the screen and the position of such
+elements may change depending on the capabilities of the display. Many game
+engines provide a way to create overlay UI elements that scale in some way to
+the space they're in using a separate UI system. In Indigo we're exporting to
+JavaScript so it makes sense for us to use other web technologies for a problem
+that is already, ultimately, solved.
```
davesmith00000, line 17: > In Indigo we're exporting to JavaScript so it makes sense for us to use other web technologies for a problem
that is already, ultimately, solved.
Sounds little like a recommended solution, I'd rather it read in such a way that we're showing people their options, maybe:

"In Indigo we're exporting to JavaScript and a browser (even in Electron), so depending on how we want the UI to look, one option is to employ more standard web technologies to do this work for us, i.e. HTML and CSS."

Or something?

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+
+Using Tyrian we can utilise HTML and CSS to ensure our overlay UI scales
+no matter what screen size we use.
+
+In this guide we'll take the basic `hello-indigo` example and add a button and
+a counter that scales with the size of the screen. It's a pretty basic example to
+be honest, but it should give you an idea of what can be achieved.
```
davesmith00000, line 24: Remove "to be honest", no need to apologise.

indigo/docs/guides/howto-hello-indigo-tyrian.md
davesmith00000: That PR is merged, and you can use links to the `.md` file and docusaurus will sort it out during the build.

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+[mill](https://github.com/PurpleKingdomGames/hello-indigo) or
+[sbt](https://github.com/PurpleKingdomGames/hello-indigo-sbt).
+In the existing implementation clicking on the dots will create a new dot
+that then rotates around the canvas. We'll modify this so that clicking a button
+does this job instead.
+
+You'll want to upgrade `hello-indigo` to use `IndigoGame`. To do this either
+follow [this guide]()
+or replace `HelloIndigo.scala` with
+[this](https://gist.github.com/hobnob/c24f00936e91a7b7e5d644d19e4f1b32)
```
davesmith00000, line 40: Is it worth these gist's you're fond of living on the pkg account? (if that's a thing?) I'm not concerned, just asking.

indigo/docs/guides/howto-hello-indigo-tyrian.md
davesmith00000: It's the same two lines, don't worry about the `val tyrianVersion = "0.3.0"` thing you have, just bake the value in.
https://tyrian.indigoengine.io/installation/#sbt

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+would usually run Electron.
+
+### Setting up the HTML and ParcelJS
+
+Usually Indigo will generate all of our HTML for us and run it through Electron,
+but for this example we're going to generate our own HTML so that we can
+inject CSS.
+
+Firstly we'll create an `index.html` and `app.js` file that looks like
+[this](https://gist.github.com/hobnob/eaa03bfe2da14562e5f2819078f16d63). You'll
```
davesmith00000, line 101: I think you should inline these here too. They're small.

davesmith00000: Also you need to remove this bit no?
```
<script>
    TyrianApp.launch("myapp");
</script>
```

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+notice we're using the direct JS output from our build here, which may feel odd.
+What will happen when we run ParcelJS through Yarn is that the HTML will be
+copied to a build directory along with the JS, CSS and any dependant static
+files (such as images) that may be needed.
```
davesmith00000, line 105: "that may be needed, and everything will be correctly linked by Parcel." ?

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+  "staticPath": "assets",
+  "staticOutPath": "assets",
+  "watcherGlob": "**"
+}
+```
+
+We need to add a couple of dependencies. Firstly we'll add the
+dependency for static files by running
+`yarn add parcel-reporter-static-files-copy --dev` from the command line; and
+secondly we'll add `snabbdom` for Tyrian by running `yarn add snabbdom`.
```
davesmith00000, line 133: I think it would be better to remove the snabbdom bit here, do all the parcel static install + config, then add snabbdom as a separate final state. It'll be cleaner for the reader and you'll get a chance to add one line explaining what snabbdom is/is for.

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+that is already, ultimately, solved.
+
+Using Tyrian we can utilise HTML and CSS to ensure our overlay UI scales
+no matter what screen size we use.
+
+In this guide we'll take the basic `hello-indigo` example and add a button and
+a counter that scales with the size of the screen. It's a pretty basic example to
+be honest, but it should give you an idea of what can be achieved.
+
+## Setup
```
davesmith00000, line 26: There's a lot of fiddly detail in this write up - which is great, necessary - but it would be good to modify the narrative arc to make it smoother and easier to follow.

To do that, I'd love to see an "Overview" / "Here's the plan" / "This is what we're going to do" section before this setup block. Write out bullets of what needs to happen and I think in doing that exercise you may be able to reorder the things to avoid the reader jumping about too much.

Or maybe not, maybe it's just difficult! But worst case, at least the reader knows what's coming up.

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+```json
+{
+  "extends": ["@parcel/config-default"],
+  "reporters":  ["...", "parcel-reporter-static-files-copy"]
+}
+```
+
+### Update HelloIndigo.scala
+
+We'll be using the `TyrianSubSystem` as a way of communicating between Indigo
```
davesmith00000, line 147: Might be good to add reciprocal links to and from here to the Tyrian page that talks about this.

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+and Tyrian. Add `import tyrian.TyrianSubSystem` to the imports in
+`HelloIndigo.scala`, and then update the `object` to be a `case class` so that
+it it contains the subsystem as an argument. We can also remove the top level
```
davesmith00000, line 150: >it it contains the subsystem as an argument
"we can pass in the subsystem as an argument"

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+     )
+```
+
+### Create a Tyrian App
+
+We'll now develop our initial Tyrian app, which will consist simply of
+a canvas (for Indigo), a counter, and a button, wrapped in a few `div` elements.
+Create a new file called `HelloTyrian.scala` inside the `helloindigo/src` folder
+and add the following contents found
+[here](https://gist.github.com/hobnob/436318b3ae5eed5891ba2b18bb8c264b).
```
davesmith00000, line 181: Again, just put it right here, I'd like to read it in context?

davesmith00000: And give some commentary.

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
++    case AddDot      extends Msg
+```.
+
+Next we'll need to deal with that message in Tyrian, by adding a case to `update`
+like this:
+
+```diff
+     msg match
++      case Msg.AddDot =>
++        (model, model.bridge.publish(IndigoGameId(gameDivId), 0))
```
davesmith00000, line 207: Technically, you don't need the `IndigoGameId(gameDivId)` stuff since you only have one game - omitting and ID means "publish to all". Might be cleaner without it?

indigo/docs/guides/howto-hello-indigo-tyrian.md
davesmith00000: Another advantage, is that we can use Mdoc to check it is correct, you just need to fence the code with `scala mdoc:silent`.

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+       val adjustedPosition = pt - model.center
+```
+
+We've also added a line below that to set a `pt` variable as we now longer have
+a mouse position. For now, we've set that to 100,100 like so which is just a
+fixed point for the dots to rotate around.
+
+Once more build the project in mill or sbt and then run `yarn start`. This time
+you'll notice that clicking the game doesn't do anything, but clicking the
+button adds a new rotating dot.
```
davesmith00000, line 240: Minor thing: If you just `yarn start` somewhere, then doing the scala build in a different terminal tab will do a hot-reload?

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
++  val init: TyrianModel =
++    TyrianModel(TyrianIndigoBridge(), 0)
+```
+
+We'll need to display that count on the website. To do this, we simply modify our
+counting `div` to display what's in the `model` like so
+
+```diff
+-      div(`class` := "counter")(),
++      div(``class`` := "counter")(model.count.toString)
```
davesmith00000, line 265: Do you have double back ticks? Might be better to use one of the aliases? Valid options:

- ``class``
- `cls`
- `className`
- `_class`

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,418 @@
...
+       case Msg.StartIndigo =>
+```
+
+For the Indigo side we need to modify the `updateModel` so that the new model is
+assigned to a variable for later use. We can then add a global event through
+the Tyrian Subsystem letting the website know how many dots we have. This is done
+with the following:
+
+```diff
+     case tyrianSubSystem.TyrianEvent.Receive(msg) =>
```
davesmith00000, line 310: `msg` is never used, which is a shame. Could you send over the x and y for the point or something?

davesmith00000 (commented):
I've added quite a few comments for improvements, but this is an excellent walk through quite a fiddly subject.

indigo/docs/guides/howto-hello-indigo-tyrian.md
hobnob: Unfortunately we can't have them on the PKG account 🙁  Worth keeping in mind though, just in case they add it

indigo/docs/guides/howto-hello-indigo-tyrian.md
hobnob: So just `model.bridge.publish(0)`?

indigo/docs/guides/howto-hello-indigo-tyrian.md
hobnob: Yup, but I think it's easier to follow as a tutorial if you're following steps, rather than switching tabs. Might be just personal preference... I just find it easier

indigo/docs/guides/howto-hello-indigo-tyrian.md
hobnob: The double backticks are to escape it for markdown - this will render as
```scala
``class``
```

hobnob: (or at least it does in my Markdown previewer)

indigo/docs/guides/howto-hello-indigo-tyrian.md
hobnob: Maybe, but ultimately it's the page X and Y, it's just simpler I think to ignore it?

indigo/docs/guides/howto-hello-indigo-tyrian.md
hobnob: Makes sense

indigo/docs/guides/howto-hello-indigo-tyrian.md
```diff
@@ -0,0 +1,445 @@
...
+You'll notice we're using the direct JS output from our build here, which may
+feel odd. What will happen when we run ParcelJS through Yarn is that the HTML
+will be copied to a build directory along with the JS, CSS and any dependant
+static files (such as images) that may be needed.
+
+Now generate an empty `css/main.css`.
+
+Next we'll need to update the `package.json` so that ParcelJS will run when yarn
+starts and copies the static files for Indigo to use. To do this add the
+following to the top of `package.json`:
```
sjrd, line 138: What is the purpose of ParcelJS?

Is it only to copy static files from one place to another? At least in sbt this would idiomatically be done as part of the build, with a call to `sbt.IO.copyDirectory` in a task.

davesmith00000: In most of the Tyrian examples, Parcel.js is used as our bundler of choice (over something like Webpack). Superficially here we're just using it to do the asset copy true, but the expectation is that we're going to use it in the development of our site, linked up resources, bundling up a deployable site, running a dev-server with hot-reloading etc.
In the server-based examples we use scalajs-bundler instead, because the Http4s service build plays nicer with that bundler in that scenario.

davesmith00000 (approved):

PR #324, opened 2022-02-05
Faster scene event processing

davesmith00000:
Plan of attack was to look at the use of `EventHandler`, wherein currently, every single node on the screen that _can_ process is events is given the opportunity to - whether it's going to or not - and that means performing a potentially expensive boundary detection on many items (in a large scene) for no reason.

As a warm up to that though, I noticed two things:

1. The entire scene graph was actually being traversed ***twice***, first by `SceneGraphViewEvents` and then by the rendering process.
2. During the first traversal is was doing that in the slowest form possible, i.e. as a `List`.

So the first commit on this branch fixes both of those things. Scene nodes are only traversed once and in the form of a `js.Array`. Good start.

For bonus points, I realised the old `SceneGraphNode` is now redundant, so I've removed it.


davesmith00000:
Latest push doesn't make event handling any prettier, but it does prevent SceneNodes trying to process events by default (faster!), and at the same time brings event handling to all primitive types (not just Sprite and Text). However I haven't yet fully tested all the primitives with event handling.

davesmith00000:
Can the event handler signature just be `GlobalEvent => GlobalEvent`?

We're always setting it up in the context of our primitive, so we can always get the bounds, and that means we never incur that cost unless we care about it.

Does that work?

indigo/indigo/src/main/scala/indigo/shared/scenegraph/Graphic.scala
```diff
@@ -98,10 +104,19 @@ final case class Graphic[M <: Material](
   def toShaderData: ShaderData =
     material.toShaderData
 
+  def onEvent(e: ((Rectangle, GlobalEvent)) => List[GlobalEvent]): Graphic[M] =
+    this.copy(eventHandler = e, eventHandlerEnabled = true)
+  def enableEvents: Graphic[M] =
+    this.copy(eventHandlerEnabled = true)
```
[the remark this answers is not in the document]
davesmith00000: Yes, in this version. My expectation is that most people will just add a handler, which will enabled the handler, and that will be the end of it for them. disable and enable I image will be the exceptional usage rather than the norm.

It does feel clunky though so I'm still thinking about it - just getting something working at the moment. More familiar semantics would be adding and removing listeners (or things that looked like listeners) but that means even more complications to the primitive scene node types.

indigo/indigo/src/main/scala/indigo/shared/scenegraph/Graphic.scala
```diff
@@ -98,10 +104,19 @@ final case class Graphic[M <: Material](
...
+  def disableEvents: Graphic[M] =
+    this.copy(eventHandlerEnabled = false)
+
 object Graphic:
 
   def apply[M <: Material](x: Int, y: Int, width: Int, height: Int, depth: Int, material: M): Graphic[M] =
     Graphic(
+      eventHandlerEnabled = false,
+      eventHandler = (_: (Rectangle, GlobalEvent)) => Nil,
```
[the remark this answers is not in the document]
davesmith00000: Nice! I did not know about that. Thanks for the tip!

davesmith00000:
> Can the event handler signature just be `GlobalEvent => GlobalEvent`?
No.

> We're always setting it up in the context of our primitive, so we can always get the bounds, and that means we never incur that cost unless we care about it.
This is not true, we're setting up _outside_ the context of the primitive, e.g. `text.onEvent(...)`.

> Does that work?
Reader, it does not work.

For two reasons:

1. You can do `PartialFunction[GlobalEvent, GlobalEvent]` but sooner or later you really need `GlobalEvent => Option[GlobalEvent]`.
2. Actually that isn't enough either though, what you really want is `SceneNode => GlobalEvent => Option[GlobalEvent]`

davesmith00000:
Update:

We're now at the point of having a vastly simpler event handler, and in the process have dropped the hard, weird old requirement for scene nodes to have a `calculateBounds` method.

This is also much faster than it used to be. That's the good news.

The bad news is that the next piece of work is to see if this horror show can be improved upon:

```scala
  def uiLayer(inputState: InputState, bl: BoundaryLocator): List[SceneNode] =
    val right = Text("AB!\n!C", 200, 2, 5, Fonts.fontKey, SandboxAssets.fontMaterial.withAlpha(0.5)).alignRight

    List(
      ...
      right.withEventHandler {
        case MouseEvent.Click(pt) if inputState.mouse.wasMouseClickedWithin(bl.findBounds(right).getOrElse(Rectangle.zero)) =>
          println("Clicked me!")
          None

        case _ =>
          None
      }
    )
```

davesmith00000:
Update, we now have something like the desired syntax improvement over the previous version:

```scala
  def uiLayer(bl: BoundaryLocator): List[SceneNode] =
    List(
      ...
      Text("AB!\n!C", 200, 2, 5, Fonts.fontKey, SandboxAssets.fontMaterial.withAlpha(0.5)).alignRight
        .withEventHandler {
          case (txt, MouseEvent.Click(pt)) if bl.bounds(txt).contains(pt) =>
            println("Clicked me!")
            None

          case _ =>
            None
        }
    )
```

indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
```diff
@@ -146,24 +152,50 @@ final class DisplayObjectConversions(
...
+        case n: DependentNode[_] =>
+          val nn = n.asInstanceOf[n.Out]
+          if n.eventHandlerEnabled then
+            inputEvents.foreach { e =>
+              n.eventHandler((nn, e)).foreach { ee =>
+                sendEvent(ee)
+              }
+            }
+
+      f(node)
```
bilki, line 196: Might be useful to rename `f` to something like `toDisplayObject`.

indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
```diff
@@ -146,24 +152,50 @@ final class DisplayObjectConversions(
...
         renderingTechnology,
-        maxBatchSize
+        maxBatchSize,
+        inputEvents,
+        sendEvent
       )
-    val l = sceneNodes.toJSArray.map(f)
-    (l.map(_._1), l.foldLeft(scalajs.js.Array[(CloneId, DisplayObject)]())(_ ++ _._2))
+
+    val l = sceneNodes.map { node =>
```
bilki, line 176: To further optimize this, I think you could `foldLeft` here instead of mapping and then folding later.

indigo/indigo/src/main/scala/indigo/shared/platform/SceneProcessor.scala
```diff
@@ -102,11 +104,13 @@ final class SceneProcessor(
           else cloneBlankToDisplayObject(blank)
 
         maybeDO match
-          case None                => acc
-          case Some(displayObject) => acc + (blank.id -> displayObject)
+          case None => acc
+          case Some(displayObject) =>
+            acc.put(blank.id.toString, displayObject)
```
bilki, line 109: I guess that this `Dictionary` acc is mutable.

indigo/indigo/src/main/scala/indigo/shared/platform/SceneProcessor.scala
```diff
@@ -143,9 +149,12 @@ final class SceneProcessor(
 
     val sceneBlend = scene.blendMaterial.getOrElse(BlendMaterial.Normal).toShaderData
 
+    displayObjectConverter.purgeEachFrame()
```
bilki, line 152: Looking at the `QuickCache` code, I'm now wondering, isn't it faster to make `cache` inside that class a `var`, then just simply drop the reference to that object, and create an empty one? Depending on the number of cached elements, it seems that purging it iterates over the entire set of keys removing them...

indigo/sandbox/src/main/scala/com/example/sandbox/SandboxView.scala
```diff
@@ -88,21 +93,20 @@ object SandboxView {
...
-      Text("AB!\n!C", 200, 2, 5, Fonts.fontKey, SandboxAssets.fontMaterial.withAlpha(0.5)).alignRight.onEvent {
-        case (bounds, _: MouseEvent.Click) =>
-          if (inputState.mouse.wasMouseClickedWithin(bounds))
-            println("Hit me!")
-          Nil
+      Text("AB!\n!C", 200, 2, 5, Fonts.fontKey, SandboxAssets.fontMaterial.withAlpha(0.5)).alignRight
+        .withEventHandler {
+          case (txt, MouseEvent.Click(pt)) if bl.bounds(txt).contains(pt) =>
+            println("Clicked me!")
+            None
```
bilki, line 107: Nice 😄

indigo/indigo/src/main/scala/indigo/shared/AnimationsRegister.scala
```diff
@@ -8,29 +8,32 @@ import indigo.shared.animation.AnimationRef
 import indigo.shared.datatypes.BindingKey
 import indigo.shared.time.GameTime
 
-import scala.collection.mutable
+final class AnimationsRegister:
```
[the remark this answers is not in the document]
davesmith00000: Yeah it's a fair question. They were at one point but I found their lifecycle easier to manage when it was an instance. But maybe it would be better as an object.

indigo/indigo/src/main/scala/indigo/shared/BoundaryLocator.scala
```diff
@@ -80,42 +85,40 @@ final class BoundaryLocator(
...
+            (acc, findBounds(node)) match
+              case (Some(a), Some(b)) => Option(Rectangle.expandToInclude(a, b))
+              case (r @ Some(_), _)   => r
+              case (_, r @ Some(_))   => r
+              case (r, _)             => r
+          }.map(_.moveBy(group.position))
+            .getOrElse(Rectangle.zero)
+      }
+
+    BoundaryLocator.findBounds(group, rect.position, rect.size, group.ref)
```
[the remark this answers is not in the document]
davesmith00000: I'm not in the habit of doing that. Maybe because I use a rainbow indent plugin in my editor (probably designed for the yaml crowd) I don't have a problem reading this stuff. It is more difficult in a PR.

indigo/indigo/src/main/scala/indigo/shared/platform/SceneProcessor.scala
davesmith00000: Good question. This cache here is a "per frame" cache, so the number of entries is likely to be low. Thus we're trading off iterating and removing vs reallocation+GC. I couldn't say which is better, maybe worth trying to benchmark it though (benchmarks are set up, just needs a new test).

indigo/sandbox/src/main/scala/com/example/sandbox/SandboxView.scala
davesmith00000: There is another form using `onEvent` which is a partial function from `(node, GlobalEvent) => GlobalEvent` that I hope will be used more as it's even cleaner. But if you want to result in no events and just `println` (which is only a debug use case really) then you do it this way.

indigo/indigo/src/main/scala/indigo/shared/platform/DisplayObjectConversions.scala
davesmith00000: I think you're right - equally I'm disinclined to get into it right now - please feel free to have a go! (...and if you don't I'll try and do so next time I'm in the area.)

PR #328, opened 2022-02-06
Fix typos in docs

Sintrastes:
This patch fixes some typos I came across while reading the docs.

### DIFF
--- a/indigo/docs/guides/howto-custom-entity.md
+++ b/indigo/docs/guides/howto-custom-entity.md
@@ -96,7 +96,7 @@ Walking through it, `position` and `depth` have been moved to be case class argu
 
 Then things get a little more complicated. `toShaderData` has been implemented by creating a `ShaderData` instance that only takes a `ShaderId`, which it sources from the companion object.
 
-The companion object defines the shader that is going to color in our entity by declaring an `EntityShader` (as a pose to a `BlendShader`). `Shader` instances whether a `BlendShader` or an `EntityShader` come in two forms: `EnityShader.Source(..)` or `EnityShader.External(..)`. The difference is that `Source` instances literally take the GLSL program as a `String`, and `External` instances look for loaded assets, which is a much nicer editing experience.
+The companion object defines the shader that is going to color in our entity by declaring an `EntityShader` (as opposed to a `BlendShader`). `Shader` instances whether a `BlendShader` or an `EntityShader` come in two forms: `EnityShader.Source(..)` or `EnityShader.External(..)`. The difference is that `Source` instances literally take the GLSL program as a `String`, and `External` instances look for loaded assets, which is a much nicer editing experience.
 
 There are several parts to an Indigo shader program that can be set (or default implementations are used), but in this case we're only setting the fragment program as follows:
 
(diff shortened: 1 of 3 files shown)


bilki (approved):
👍

davesmith00000:
Well! This has led me on a merry trip round the internet.

First of all, yes, I have been making a typo presumably forever: "a pose" should be "appose". Thank you for forcing me to looking into it. 😄 

"oppose" vs "appose" - I've been using appose my whole life out of habit without ever checking the definition.

Many people on the internet believe that appose is old fashioned, and that oppose is common now (possibly more so in American english?). But what do people on the internet know? Let's check the dictionary:

- *Oppose* - to act against or provide resistance to
- *Appose* - to place side by side, as two things; place next to; *juxtapose*.
- *Juxtapose* - to place close together or side by side, especially with an arresting or surprising effect, or in a way that invites comparison or contrast.

So while there certainly is a typo there, I think I really did mean "appose". I am fully open to correction though if you have an ...opposing... (sorry) argument.

Sintrastes:
> 

@davesmith00000 I might be wrong, but as far as I can tell, whereas "appose" is a valid word in both British and American English (might be more common to use in British English, not sure), they are not synonyms in either dialect.

As I understand it, "appose" means something like "alongside of", where as "oppose" means something like "contracted with".

Again, correct me if I'm wrong (I didn't even know appose was a proper word before looking it up! So definitely not common in American English), but as far as I can tell (with a quick Google search), "as apposed to" is not really a valid idiom in either British or American English. The idiom in both is "as opposed to".

Sintrastes:
Funny story, I just ran into the difference between "canceled" (American spelling) and "cancelled" (everyone else) the other day. Was trying to figure out why my IDE was not finding "cancellableSeq", when I knew I had a function of that name.

Turns out I had named it before "cancelableSeq".

I blame Webster. 😄

Sintrastes:
> Google is _BIAS_ against appose!! 😄
> 
> But if you force google not to correct you into "oppose", you get gems like this:
> 
> image (Australian)
> 
> image (Welsh)
> 
> image (British)
> 
> I care about this less than I'm arguing at this point. Just interested.

Same here. No bike shedding necessary. 😄 

But, out of interest, I decided to dig into this a bit more.

First off, I now think that "as apposed to" does make semantic sense, and is synonymous with "as opposed to", since appose means "place (something) in proximity to or juxtaposition with something else" -- so it could be putting two things in proximity to contrast them. 

However, looking at [google trends](https://trends.google.com/trends/explore?q=as%20opposed%20to,as%20apposed%20to) worldwide (at least if we assume that google trends is a fair estimator of usage for phrases like this), it looks like worldwide "as opposed to" is by far the more common expression -- and is even in the UK much more common.

If this were just a US v.s. UK thing (or especially if it were a US the rest rest of the English speaking world thing!), I'd be fine changing this to "apposed", but I think that given this data, for readability purposes, it seems like "opposed" would be the more understandable option for English speakers worldwide.

Of course, all up to you @davesmith00000!

davesmith00000:
Well! Who am I to contradict global word usage trends! 😂

I shall keep "apposed" (spelling corrected) for personal use and switch to "opposed" for general consumption.

Thank you for joining me on this plunge into an obscure rabbit hole. I'm not usually this pedantic I promise. 😁

Will merge soon, I have a few docs PRs to sort out.

PR #332, opened 2022-02-12
Fixed #331: Option to disable frame rate

### LINKED ISSUE #331: Add option to disable frame rate throttling

davesmith00000:
Adds two flags:

1. A plugin flag to disable the frame rate limit (in electron) means your game with run as fast as it possibly can.
2. A config flag that moves the long standing "frame rate" to an optional "frame rate limit".

The recommendation is to let the browser manage the frame rate unless you specifically want to go slower. To do that, you set the plugin flag to `false` and the frame rate limit config flag to `None`, and these are the defaults.

### DIFF
--- a/indigo/indigo-extras/src/main/scala/indigoextras/subsystems/FPSCounter.scala
+++ b/indigo/indigo-extras/src/main/scala/indigoextras/subsystems/FPSCounter.scala
@@ -22,7 +22,7 @@ import indigo.shared.time.Seconds
 
 final case class FPSCounter(
     startPosition: Point,
-    targetFPS: FPS,
+    targetFPS: Option[FPS],
     layerKey: Option[BindingKey],
     fontFamily: FontFamily,
     fontSize: Pixels
@@ -42,15 +42,21 @@ final case class FPSCounter(
   def update(context: SubSystemFrameContext, model: FPSCounterState): GlobalEvent => Outcome[FPSCounterState] = {
     case FrameTick =>
       if (context.gameTime.running >= (model.lastInterval + Seconds(1)))
+        val nextFps = targetFPS match
+          case None =>
+            model.frameCountSinceInterval + 1
+
+          case Some(target) =>
+            Math.min(target.toInt, model.frameCountSinceInterval + 1)
+
         Outcome(
           model.copy(
-            fps = Math.min(targetFPS.toInt, model.frameCountSinceInterval + 1),
+            fps = nextFps,
             lastInterval = context.gameTime.running,
             frameCountSinceInterval = 0
           )
         )
-      else
-        Outcome(model.copy(frameCountSinceInterval = model.frameCountSinceInterval + 1))
+      else Outcome(model.copy(frameCountSinceInterval = model.frameCountSinceInterval + 1))
 
     case FPSCounter.Move(to) =>
       Outcome(model.copy(position = to))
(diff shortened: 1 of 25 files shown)


indigo/indigo-extras/src/main/scala/indigoextras/subsystems/FPSCounter.scala
```diff
@@ -42,15 +42,21 @@ final case class FPSCounter(
   def update(context: SubSystemFrameContext, model: FPSCounterState): GlobalEvent => Outcome[FPSCounterState] = {
     case FrameTick =>
       if (context.gameTime.running >= (model.lastInterval + Seconds(1)))
```
bilki, line 44: Since `targetFPS` value won't change, this could be a curried function outside of `update`.

indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
```diff
@@ -43,16 +43,22 @@ final class GameLoop[StartUpData, GameModel, ViewModel](
     _runningTimeReference = time
     val timeDelta: Long = time - lastUpdateTime
 
-    if timeDelta >= gameConfig.frameRateDeltaMillis.toLong - 1 then
-      runFrame(time, timeDelta)
-      gameEngine.platform.tick(gameEngine.gameLoop(time))
-    else gameEngine.platform.tick(loop(lastUpdateTime))
+    gameConfig.frameRateLimit match
```
bilki, line 46: Similar to the comment above, a curried function branching upfront?

indigo/indigo/src/main/scala/indigo/shared/time/FPS.scala
```diff
@@ -16,3 +16,5 @@ object FPS:
     def toDouble: Double   = fps.toDouble
     def toSeconds: Seconds = Seconds(1.0d / fps.toDouble)
     def toMillis: Millis   = toSeconds.toMillis
+
+  given CanEqual[FPS, FPS] = CanEqual.derived
```
bilki, line 20: `derives CanEqual` should give you this instance for free 🤔

indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
davesmith00000: So I didn't need a curried function since the values are totally fixed, but deciding the branches once ahead of time is a very sneaky idea and it working nicely - thanks!

```scala
  private val runner: (Long, Long, Long) => Unit =
    gameConfig.frameRateLimit match
      case None =>
        (time, timeDelta, lastUpdateTime) =>
          runFrame(time, timeDelta)
          gameEngine.platform.tick(gameEngine.gameLoop(time))

      case Some(fps) =>
        (time, timeDelta, lastUpdateTime) =>
          if timeDelta >= gameConfig.frameRateDeltaMillis.toLong - 1 then
            runFrame(time, timeDelta)
            gameEngine.platform.tick(gameEngine.gameLoop(time))
          else gameEngine.platform.tick(loop(lastUpdateTime))

  def loop(lastUpdateTime: Long): Long => Unit = { time =>
    _runningTimeReference = time
    val timeDelta: Long = time - lastUpdateTime
    runner(time, timeDelta, lastUpdateTime)
  }
```

indigo/indigo/src/main/scala/indigo/shared/time/FPS.scala
davesmith00000: On an opaque type?

bilki: Ah, didn't notice the opaque type, nevermind then 😶‍🌫️

bilki (approved):
Great!

PR #333, opened 2022-02-12
Fixed #327: Dice roll from zero bug

### LINKED ISSUE #327: Dice rolling from zero not working as expected
Currently, rolling a dice works as follows:
```
  def roll(sides: Int): Int =
    r.nextInt(sanitise(sides)) + 1

  def rollFromZero(sides: Int): Int =
    roll(sides + 1) - 1
```
This means that rolling a 6-sided dice with `roll(6)` will output a value in the range `[1, 6]`, as intended and stated by the `Dice` docs.

The problem comes when calling `rollFromZero(6)`, which expands to:
```
rollFromZero(6) ==
roll(6 + 1) - 1 ==
r.nextInt(6 + 1) + 1 - 1 ==
r.nextInt(7) ==
value in [0, 7)
```
That could result in a value `6`, when the expected range is `[0, 5]`, as the dice contains 6 sides.

As part of this issue, it might be a good idea to add some scalacheck tests to test the `Dice` class and functions.

davesmith00000:
The bug has been fixed.

While I was there I removed `arbitrary` on the basis that is was unspeakably weird, and renamed and hide some things for clarity. I also added a `shuffle` function... because that seemed like a a good idea.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/dice/Dice.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/dice/Dice.scala
@@ -57,7 +57,7 @@ object Dice:
       rec(numberOfDice - 1, NonEmptyList(diceSidesN(sides, seed).roll))
     else None
 
-  private[indigo] def loaded(fixedTo: Int): Dice =
+  def loaded(fixedTo: Int): Dice =
     new Dice {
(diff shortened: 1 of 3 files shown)


bilki (approved):
Looking good 👍

PR #348, opened 2022-03-14
Override Key.hashCode to be compatible with equals.

### LINKED ISSUE #347: indigo.shared.constants.Key overrides equals but not hashCode
equals is case-insensitive, but hashCode is not, so
```scala
Key(74, "j") == Key(74, "J")
```
but
```scala
!Set(Key(74, "j")).contains(Key(74, "J"))
```

tsnee:
fixes #347

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/constants/Key.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/constants/Key.scala
@@ -20,6 +20,7 @@ final case class Key(code: Int, key: String) derives CanEqual {
       case _ => false
     }
 
+  override def hashCode: Int = code
 }
 
 object Key {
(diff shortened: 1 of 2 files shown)


davesmith00000 (approved):
I'm not 100% that's the right thing for Key, but at least it's consistent. Thanks! 😄

tsnee:
Glad to be of service.

I also have some generic tileset/tilemap/sprite sheet code I could contribute. One problem with it is that it takes a configuration object that requires the dimensions of the graphic file. I believe that information should be figured out by loading the file and examinging it directly. I'm sure Indigo has code that can get the dimensions of an image, but I don't know where to find it.

tsnee:
I accidentally added a new commit, which undid your review. Reverting the commit did not help. Sorry about that. I will clean up the Git history and force-push something cleaner.

davesmith00000 (approved):

PR #349, opened 2022-03-15
Small rectangles now overlap correctly

hobnob:
Previously small rectangles would sometimes report that they weren't overlapping, when in fat they were. This PR fixes that by making use of half sizes that are doubles rather than ints


davesmith00000 (approved):
LGTM! 😄

PR #350, opened 2022-03-15
Add utility for generic sprite sheet format.

tsnee:


indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
...
+/** Utility for working with generic sprite sheets. The
+  * sheet must be uniform, with each sprite having the same
+  * width and height, and a (possibly zero-pixel) margin.
+  */
+final case class GenericTileset(
+    assetName: AssetName,
+    width: Int,
+    height: Int,
+    tileWidth: Int,
+    tileHeight: Int,
```
davesmith00000, line 14: The convention is that the main constructor will use `Size` instead of separate width and height values. What we then do is has a companion with an apply method that gives additional (more loosely typed) construction options. So this would be (perhaps):

```scala
assetName: AssetName, imageSize: Size, tileSize: Size, margin: Int
```

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
+package indigo.shared.formats
+
+import indigo._
+
+/** Utility for working with generic sprite sheets. The
+  * sheet must be uniform, with each sprite having the same
+  * width and height, and a (possibly zero-pixel) margin.
+  */
+final case class GenericTileset(
```
davesmith00000, line 9: We could just call it `Tileset`?

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
+package indigo.shared.formats
```
davesmith00000, line 1: An unexpected but good choice for the package. I was wondering where it should go after you suggested this feature! After you've looked into the comments below, please add an alias to the `package.scala` (copy the type+val pattern in there).

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
...
+) {
+  private val rectangles =
+    for {
+      y <- 0 until height by tileHeight + margin
+      x <- 0 until width by tileWidth + margin
+      rectangle = Rectangle(x, y, tileWidth, tileHeight)
+    } yield rectangle
+
+  protected[formats] def lookupRectangle(idx: Int): Option[Rectangle] =
+    rectangles.lift(idx)
```
davesmith00000, line 25: I don't think this needs to be private/protected. Why would someone wish to look up the rectangle? Who knows! But it's a valid operation on the type and it's self contained so...

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
...
+  * and increases as it goes right and down
+  * @return an optional Graphic, None if the given index was out of bounds.
+  */
+  def apply(idx: Int): Option[Graphic[Material.Bitmap]] =
+    lookupRectangle(idx).map { rectangle =>
+      Graphic(
+        Rectangle(0, 0, tileWidth, tileHeight),
+        Material.Bitmap(assetName)
+      )
+        .withCrop(rectangle)
```
davesmith00000, line 37: Might be better to use one of the other constructors to avoid the `.withCrop` since that causes an allocation. In your own code that's totally fine, but in the engine code we should keep it lean where we can.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
...
+    tileWidth: Int,
+    tileHeight: Int,
+    margin: Int
+) {
+  private val rectangles =
+    for {
+      y <- 0 until height by tileHeight + margin
+      x <- 0 until width by tileWidth + margin
+      rectangle = Rectangle(x, y, tileWidth, tileHeight)
+    } yield rectangle
```
davesmith00000, line 22: So this works, but you can just do this with math. You don't even need the image height:
https://github.com/PurpleKingdomGames/roguelike-starterkit/blob/701c3e2f7ffb027fae694ea900a29708ab5ffd85/roguelike-starterkit/src/main/scala/io/indigoengine/roguelike/starterkit/utils/PathFinder.scala#L160

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  protected[formats] def lookupRectangle(idx: Int): Option[Rectangle] =
+    rectangles.lift(idx)
+
+/** @param idx a zero-based index that begins at the upper right of the sheet
+  * and increases as it goes right and down
+  * @return an optional Graphic, None if the given index was out of bounds.
+  */
+  def apply(idx: Int): Option[Graphic[Material.Bitmap]] =
```
davesmith00000, line 31: I think this apply method is a good idea - I wouldn't have thought of doing it this way - but it's good. Might be nice to offer another version more explicitly named like `fromIndex` (helps people find what can be done via their IDE), and have this apply method delegate to it?

davesmith00000 (changes requested):
Great start. For a relatively small code block I'm afraid I have left many comments, but a good percentage of them are just style points. Thanks!

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
davesmith00000: Of course, you may want the image height for validation, and the link above would then need to be multiplied out by the size.

The advantage of your approach is that you've pre-calculated the rectangles just one, at an instantiation cost. If someone made one of these on every frame to look up a single index, they're paying a lot for little benefit. If they make one once, build a bunch of graphics ahead of time and store them and throw this instance away, then the cost is totally worth it. However, the maths calculation is cheap, so unless I've missed something (I've yet to have my first coffee of the day so... likely!) then you might as well do the math?

davesmith00000:
Follow up thought: The point of this is convenient use of tilesets so that people can focus on their game building. What other things would be helpful here that we can do for them?

```scala
val maxRows: Int = ???
val maxColumns: Int = ???
val maxIndex: Int = ???
val length: Int = maxIndex

def fromCoords/Point(pt: Point): Option[Graphic[Material.Bitmap]] = ???
```

Anything else you can think of?

On that last one, lots of excellent reasons to need to look up by index, but people are bad at indexes, and they might also like to just be able to look up by coordinates now and again.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: So, move the current argument list to an apply method? OK.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: Sure.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: OK. While I'm at it, I'll add them for Aseprite and TiledMap in the same package.

tsnee: Never mind, already there.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: OK

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: Absolutely.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: Turns out `TileSet` is already being used in TiledMap.scala. I would have thought that `Tileset` with a lowercase s would not conflict, but I get all kinds of scalajs linking errors when I change the name. What should I do?

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: I guess this was a case of premature optimization. In my own code, I instantiate the class once, but call its apply method ~240,000 times per second, so it seemed like a no-brainer.

indigo/indigo/src/main/scala/indigo/shared/formats/GenericTileset.scala
tsnee: Sure.

davesmith00000 (approved):
LGTM

Issue #362, opened 2022-04-16
(Re)Consider replacing `List` on the APIs


davesmith00000:
It's currently very easy for someone to hit the performance limits of the engine with some fairly ordinary list manipulations.

I've considered this before but rejected it on the basis that list semantics are nice and friendly, but maybe a little pain is worth it.

The properties the data structure most needs is (I think):

1. Fast construction
2. Fast traverse
3. Fast concatenation
4. Minimal allocation

@sjrd has suggested `IArray` or possibly a custom data structure built on `IArray` such as a [Rope](https://en.wikipedia.org/wiki/Rope_(data_structure)).

armanbilge:
Depending on your exact constraints ...

I wonder if this talk on `fs2.Chunk` (or `fs2.Chunk` itself) would be helpful to you.
https://youtu.be/wOybldcyMLs

There's also [`cats.data.Chain`](http://typelevel.org/cats/datatypes/chain.html).

davesmith00000:
Interesting talk, thanks for sharing, some great details and considerations there. The binary search work that Michael talks about at the end is not dissimilar to the Rope suggestion it seems to me.

`Chain` feels closer to what I'm after than `Chunk` though, someone else also kindly recommended I look into it.

The interesting thing about this structure is that (famous last words...) I need almost no manipulation operations, as long as I can construct it and traverse it efficiently, that's basically all I need. It can be rather specialised.

davesmith00000:
@armanbilge I've been having some fun in the PR attached.

davesmith00000:
Will be in the next release.

PR #368, opened 2022-05-02
Replace List with a new type focused on construction speed

davesmith00000:
Relates to issue #362

The initial thought was that this would only be for use in the `SceneUpdateFragment`, but maybe it should be more general purpose...

The idea essentially is to reduce friction from building your view to rendering, as currently it's easy to destroy performance by doing normal List-like things.

***Currently called `Chain` due to lack of inspiration. It isn't meant to be _quite_ the same as `cats.Chain`. Probably.***


davesmith00000:
Preliminary benchmark results are promising.

image

indigo/indigo/src/main/scala/indigo/shared/collections/Chain.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  final case class Combine[A](chain1: Chain[A], chain2: Chain[A]) extends Chain[A]:
+    val isEmpty: Boolean            = chain1.isEmpty && chain2.isEmpty
+    def map[B](f: A => B): Chain[B] = this.copy(chain1 = chain1.map(f), chain2 = chain2.map(f))
+    def foreach(f: A => Unit): Unit =
+      chain1.foreach(f)
+      chain2.foreach(f)
+    def head: A                        = chain1.head
+    def headOption: Option[A]          = chain1.headOption
+    def toJSArray[B >: A]: js.Array[B] = chain1.toJSArray ++ chain2.toJSArray
```
armanbilge, line 155: There's a lot more room for optimization here, right? Essentially you should be able to allocate one array, and traverse the tree structure adding directly to that array.

davesmith00000: Funnily enough, that was the original implementation, then I changed it, and then I decided to stop fiddling until I could benchmark. Yes I think there's more to do.

Also the toJSArray benchmarks are currently totally unfair I realised afterwards as they just pass the original array back.

Anyway, it's been fun so far. 🙂

armanbilge: > Also the toJSArray benchmarks are currently totally unfair I realised afterwards as they just pass the original array back.

Lol. They did look stupidly good 😆

davesmith00000: > Lol. They did look stupidly good 😆

Making stupid mistakes very publicly is my MO. Luckily I have no pride.

image

That's more like it. Now we can see there's definitely work to do.

davesmith00000: Yeah so, interesting. The next problem is that if you want to do the trick of pushing into a single pre-allocated array, then you need to know the sizes, positions and offsets of everything. This moves us into Chunk/Rope territory of needing metadata of accumulated sizes or a binary tree of lengths. Exciting stuff!

davesmith00000:
image

Significantly improved toJSArray for nested chain structure. The next trick is probably to remove the pattern match with a method that does a foreach while allowing you to know what index you're looking at.

I don't know if this is a good use of my time, but it's very interesting.

indigo/indigo/src/main/scala/indigo/shared/collections/Chain.scala
davesmith00000: > Essentially you should be able to allocate one array, and traverse the tree structure adding directly to that array.

image

For the `map` operation (not `toJSArray`): In this table we have the original, then in the second two I'm allocating to a fixed array (one where I give it the size and one where I don't bother, the former is faster). Performance is worse however.

So it looks like for traversal operations, I can't do better than just looping the structure as efficiently as possible. It will always be slower than a native `map` or `foreach` - but I presume that's down to the megamorphic dispatch and additional allocation costs. I may well be missing something though?

That's not the end of the world. This structure at the moment is meant for fast construction, which it's doing just fine.

davesmith00000: Ah, but that depends on the depth and complexity of the tree. 🤔
On a complicated tree I'd be better off either doing the big array or flattening the whole thing first.

davesmith00000:
Latest results:

image

davesmith00000:
The next job is to try and replace `List` everywhere and find out all the exciting reasons that this doesn't work as expected.

davesmith00000:
I don't think that the numbers have changed much since the last upload, but the benchmarks again:

image

There are almost certainly issues with the way I'm testing those, for example, I'm confident that the more nesting we have the worst the `toJSArray` numbers will get. However, most games I suspect don't nest enough to be a cause for concern, and will still see an improvement in general.

At this point, `Batch` is very very similar to Cats' `Chain`, the main difference is just that it's backed by `js.Array`. Building this has been very interesting, I've learned many things.

I've now replaced many instances of `List` and `Array` (where appropriate) with `Batch`, in some cases there is little obvious performance gain, in text heavy scenes (`Text` not `TextBox`) there is a modest improvement, in other cases the FPS jumped by ~40%. Results very scene dependant, but they are no worse than they were before.

The point of the `Batch` type is those two areas at the start and end of the benchmarks. Fast scene construction, and reduced friction with the rendering engine (i.e. collection type conversion cost). To that end, although it supports a lot of the normal functions you'd expect to see on a collection type, mostly it's just delegating to `js.Array`. All the work is in how it's constructed, and that bit is - as I said - almost the same as `Chain`.

davesmith00000:
Things to do:

- [_] Add `toBatch` helpers into `import indigo.syntax.*`
- [_] Upgrade the roguelike starterkit and test
- [_] Upgrade my roguelike game and try it all out in a real world scenario

Issue #369, opened 2022-05-06
Low fps when using Mutants on Windows


TiredFalcon:
When using Mutants:

- On Chrome 101 for macOS Monterey I get ~120 fps.
- On Chrome 100 for Windows 10 I get ~3 fps.

Which is a wild difference.
I tried recording with the devtools profiler (but I'm not at all experienced in using it) and the striking difference is that on Chrome for Windows, 98% of Self Time is attributed to getUniformBlockIndex.
On Chrome for macOS, the Self Time of the activities is spread more evenly.
I tried on Microsoft Edge and also there it's 3 fps.

Deactivating the Mutants resulted in high fps also on Windows.

This issue occurred with this game: https://gitlab.com/aronfiechter/space-survival-redux/-/tree/dev

davesmith00000:
Have merged in the above PR, but hopefully I'll be able to try this out on a windows machine myself in the near future, so I'll leave this open for now as a reminder.

davesmith00000:
Tested, I _think_ this is fine in so much as mutants themselves aren't a problem anymore. You can still kill performance depending on how you choose to use Mutants, but that's harder for me to do anything about.

If a new instance of this being a problem comes up, or if @TiredFalcon comes back and says there is still a problem then we'll either re-open this issue or make a new one.

PR #370, opened 2022-05-07
Improving Mutant Performance

davesmith00000:
Related to issue #369 

cc @TiredFalcon


davesmith00000:
So my opening gambit to speed things up is to address this:

> Chrome for Windows, 98% of Self Time is attributed to getUniformBlockIndex.

The change I've tried is to change when the call to getUniformBlockIndex is made. Calls to the GPU are slow, very slow. On a mac this doesn't seem to be a big problem, but in the issue report it seems windows is spending a long time her in particular when using Mutants.

Whether or not this fixes the Windows issue, this small change lets me have ~60% more mutants.

davesmith00000:
I don't know if this fixed the mutant problem reported, but it does give a performance boost, so I'm inclined to merge anyway. Will have one more go at trying to replicate the problem first though

PR #385, opened 2022-06-20
Context menu config parameter

zikolach:


indigo/indigo/src/main/scala/indigo/platform/Platform.scala
```diff
@@ -44,8 +44,8 @@ class Platform(
   val rendererInit: RendererInitialiser =
     new RendererInitialiser(gameConfig.advanced.renderingTechnology, globalEventStream, dynamicText)
 
-  @SuppressWarnings(Array("scalafix:DisableSyntax.null", "scalafix:DisableSyntax.var"))
-  private var _canvas: Canvas = null
+  @SuppressWarnings(Array("scalafix:DisableSyntax.var"))
+  private var _canvas: Canvas = _
```
davesmith00000, line 48: I don't think this is correct. `_` sets an initial value from the language default (I think?) but here that can't exist, so `null` is accurate, unless the spec says that it will be `null`... in which case why not be explicit?
Happy to be proved wrong.

davesmith00000 (commented):
I've left one comment that I'm not sure about, but otherwise looks good, are you happy that this is good to go?

Could I interest @bilki in this one too I wonder? 😄

davesmith00000:
Apologies for the radio silence, going to try and merge this in.

PR #386, opened 2022-06-20
Add ability to customize html background color

zikolach:
It is useful while developing a game with dark background (to avoid showing white empty HTML page on browser reloads).

### DIFF
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/Utils.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/Utils.scala
@@ -6,7 +6,8 @@ final case class TemplateOptions(
     title: String,
     showCursor: Boolean,
     scriptPathBase: Path,
-    gameAssetsDirectoryPath: Path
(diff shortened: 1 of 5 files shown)


zikolach:
@davesmith00000 could you please have a look? I made changes according to your suggestions. Switching white to black really hurts the eyes when reloading page during development 👓

indigo-plugin/indigo-plugin/src/indigoplugin/Utils.scala
```diff
@@ -6,7 +6,8 @@ final case class TemplateOptions(
     title: String,
     showCursor: Boolean,
     scriptPathBase: Path,
-    gameAssetsDirectoryPath: Path
+    gameAssetsDirectoryPath: Path,
+    backgroundColor: String = "white"
```
davesmith00000, line 10: Do we need the default? Can it be removed? I suspect it's unnecessary? SBT will default to `white` and Mill forces you to provide a value.

PR #389, opened 2022-07-03
Working on Batch improvements

davesmith00000:


indigo/indigo/src/main/scala/indigo/shared/collections/Batch.scala
```diff
@@ -132,7 +133,8 @@ sealed trait Batch[+A]:
     *   `String`
     */
   def mkString(separator: String): String =
-    head.toString + separator + tail.toJSArray.mkString(separator)
+    if isEmpty then ""
+    else head.toString + separator + tail.toJSArray.mkString(separator)
```
lomigmegard, line 137: suggestion: use directly the underlying implementation `this.toJSArray.mkString(separator)`, which doesn't need the `if` or the head/tail manipulation.

davesmith00000: Good shout, didn't really engage brain there. 😄

Issue #387, opened 2022-06-29
Don't re-set UBO's that haven't changed


davesmith00000:
If you had 30 objects that were all the same, each with 2 UBOs and the data in the first one never changed (e.g. static config), there should be no reason to reset those UBOs, saving data transfer /GPU calls.

PR #394, opened 2022-07-14
Do not re-set UBOs whose data has not changed

davesmith00000:
This is for issue #387, and needs further testing somehow.

The idea is simple: Setting UBO data is expensive (requires 6 or 7 calls to the GPU) and in the current version of Indigo, if any part of it has changed, we re-set the whole lot. However, UBO data is actually set in a series of chunks (structs), so you can choose to _not_ set a chunk if you're confident the data is the same, saving yourself GPU calls and data transfer costs.

...but it's really hard to test.


davesmith00000:
Following testing, I cannot currently see any drawbacks or faults introduced with this change. Merging, will keep an eye on it.

PR #398, opened 2022-08-12
Bugfix aseprite json

solarized-fox:
I encountered a couple problems with the aseprite importer. firstly, when I exported my sprite sheet's JSON, it lacked a field for `AsepriteMeta.image`, which caused the decoding process to fail. it appears that that field is [optional,](https://github.com/aseprite/aseprite/blob/e6c90334f96f9553288c8b1bea2e654238e8816a/src/app/doc_exporter.cpp#L1297) though I don't actually know exactly what triggers it. at any rate, indigo doesn't seem to use that field, so I propose removing it.

secondly, the documentation mentions that only the aseprite's hash format is supported, however the required `AsepriteFrame.filename` field seems to be only present in the array export format (that field's value is used as the key in hash exports). this leads me to  strongly suspect this was a simple typo, as importing the array JSON works fine, so I fixed the relevant documentation as well.

### DIFF
--- a/indigo/docs/platform/importers.md
+++ b/indigo/docs/platform/importers.md
@@ -21,7 +21,7 @@ Graphics can be exported as any web compatible file format. One thing to be awar
 
 Animations can also be exported as a spritesheet with accompanying JSON data, here is an example of the export window.
 
-**PLEASE NOTE:** The export is done as a "hash", Indigo does not support the other "array" option.
+**PLEASE NOTE:** The export is done as an "array", Indigo does not support the other "hash" option.
 
 ![Aseprite export dialogue window](/img/aseprite_export.png "Aseprite export dialogue window")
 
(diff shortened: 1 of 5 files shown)


indigo/docs/platform/importers.md
```diff
@@ -21,7 +21,7 @@ Graphics can be exported as any web compatible file format. One thing to be awar
 
 Animations can also be exported as a spritesheet with accompanying JSON data, here is an example of the export window.
 
-**PLEASE NOTE:** The export is done as a "hash", Indigo does not support the other "array" option.
+**PLEASE NOTE:** The export is done as an "array", Indigo does not support the other "hash" option.
```
davesmith00000, line 24: Well spotted - I think you're right!

Here is an export I use in one of the demos:
https://github.com/PurpleKingdomGames/indigo/blob/main/demos/pirate/assets/Ship%20Helm.json

And here is a json-array export example:
https://gist.github.com/dacap/a32adb9248320326733a
from their [docs](https://www.aseprite.org/docs/cli?l=You%20can%20convert%20or%20export%20your%20sprites%20to%20other%20formats%20(or%20textures%2Bjson%20data)%20from%20the%20command%20line.%20%20See%20Platform-specific%20Details%20section%20to%20know%20how%20to%20use%20the%20command%20line.%0A%0AOptions%0AUse%20Cases%0APlatform-specific%20Details%0AAutomating%20the%20process%0A%0A%0AOpt...&t=json#format).

davesmith00000 (approved):
These all look like sensible changes - thank you!

davesmith00000:
I had to approve the CI run, will merge on successful completion. Thanks again. :+1:

PR #399, opened 2022-08-12
Reset lastFrameAdvance after an AnimationAction.jumpTo

solarized-fox:
I noticed that the frame some of my animations started on seemed somewhat random, even after calling `Sprite.jumpToFirstFrame`. after doing some digging, I believe this code is the culprit. it seems the way animations in indigo work is that a timestamp is saved each time a frame advances, and when the time since that timestamp is greater than the current frame's designated duration, it flips to the next frame

if so, reusing the `lastFrameAdvance` value here seems incorrect. for example, if you call `Sprite.jumpToFirstFrame` midway through frame n, it will effectively jump to midway through frame 0 rather than the beginning, which I don't think is the expected effect. using the current time instead seems to fix the issue I was having, but it's possible there's an intentional design choice I'm not aware of, so feel free to use your discretion here

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/animation/AnimationRef.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/animation/AnimationRef.scala
@@ -118,13 +118,13 @@ final case class CycleRef(
           cycle // No op, done at animation level.
 
         case JumpToFirstFrame =>
-          updatePlayheadAndLastAdvance(0, lastFrameAdvance)
+          updatePlayheadAndLastAdvance(0, gameTime.running.toMillis)
 
         case JumpToLastFrame =>
-          updatePlayheadAndLastAdvance(frameCount - 1, lastFrameAdvance)
+          updatePlayheadAndLastAdvance(frameCount - 1, gameTime.running.toMillis)
 
         case JumpToFrame(number) =>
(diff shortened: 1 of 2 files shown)


davesmith00000:
I think I've seen this myself very occasionally - I suspect I just don't use them very often / in this way so the bug hasn't struck me as being a big thing. But I think you're probably right and it would be great to get it resolved!

What I don't know without working on it, is whether or not this is the right fix.

Do you think you could try and write a unit test that shows and then fixes the issue? There is already an `AnimationRefTests` test suite you could make use of.

solarized-fox:
@davesmith00000 yeah I noticed it because a jump animation I'm using had two frames which gave it the appearance of having random playback haha. it's late here but I will look into writing unit tests tomorrow, I wanted to be sure my hypothesis wasn't way off base before I put the time into writing tests. thanks for the tip on the `AnimationRef` test suite, that'll make things a lot easier than I was anticipating 👍

solarized-fox:
alright, I wrote up a test that demonstrates the problem. this should pass on this branch, but if you check this file out on main it will fail on line 107

davesmith00000 (approved):
Thank you for adding this test case and apologies for taking a while to review this. I haven't manually tested it yet, but I think this looks perfectly reasonable and I'm happy to merge it in.

Thanks again!

PR #402, opened 2022-08-26
example mill template in docs is incomplete

solarized-fox:
the "minimal build.sc" mentioned in the docs won't actually compile (even when actual version numbers are substituted in, that is) as it is missing values for two members. the solution was to copy the two missing lines from [here](https://github.com/PurpleKingdomGames/blank-indigo-mill/blob/f4aa935fe88e8f6455bcc8f4390c41c40fecfc97/build.sc). if these are not sensible defaults feel free to say so,  but I felt it was worth bringing to attention as it seems unintentional


davesmith00000 (approved):
Thank you very much! :smile:

PR #403, opened 2022-08-28
simplification: replace SceneAudio.None with Option types

solarized-fox:
this PR is the result of a rabbit hole I went down after trying to mute a `SceneUpdateFragment` by `|+|`-ing it with a `SceneUpdateFragment(myLayer).withAudio(SceneAudio.None)` and being surprised when music kept playing. I ended up figuring out that `SceneAudioSource.None` is a special value that represents both silence _and_ the absence of a value, that is to say, it plays no sound, but also any other `SceneAudioSource` will override it when combined.

this seems slightly more complex than just using `Option` types to represent sound and lack thereof, so I'm suggesting a few changes. to summarize:

* `SceneAudioSource.None` is gone, as all `SceneAudioSource` values now represent some sort of audible sound.
* `SceneAudio`'s three `SceneAudioSource` members are now of type `Option[SceneAudioSource]`.
  * `SceneAudio.Mute` has been added, which is simply `SceneAudio(None, None, None)`.
* `PlaybackPattern.Silent` has been removed, as silence is now represented by a `None` in a `SceneAudio`
* `SceneUpdateFragment.audio` is now an `Option[SceneAudio]`, which makes the empty-value semantics more obvious.
  *  while `SceneAudio.Mute` is not special in any way, it is the fallback used by `AudioPlayer` if the final `SceneUpdateFragment.audio` value passed to it ends up being a `None`.
 * for consistency, the three `AudioSourceState` vars in `AudioPlayer` are now `Option[AudioSourceState]`.
   * this eliminates the need for the `nodes` field to be an `Option`, so that was changed as well.
 
 to recap, the advantages of these changes are:
 * no more magic `SceneAudioSource(BindingKey("none"), PlaybackPattern.Silent, Volume.Min)` value.
   * silence cannot interrupt silence, so silent values probably do not need `BindingKey`s anyway?
 * using a regular old `Option` in the API leads to fewer surprises for the end user.
 * `SceneAudio.Mute` can be used both as a noun (the audio is mute) and a verb (a fragment with it will mute the left-hand-side fragment when `|+|`'d).
 * this is basically just a syntax change, the semantics are intended to be more or less identical.
 * no increase in codebase size.

feel free to give feedback or ask for tests/changes, if you don't feel this PR is a good fit, I'll understand that too. consider this just something that I thought of and decided to share in case it might be helpful!

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/platform/audio/AudioPlayer.scala
+++ b/indigo/indigo/src/main/scala/indigo/platform/audio/AudioPlayer.scala
@@ -130,13 +130,15 @@ final class AudioPlayer(context: AudioContextProxy) {
     }
 
   @SuppressWarnings(Array("scalafix:DisableSyntax.var"))
-  private var sourceA: AudioSourceState = new AudioSourceState(BindingKey("none"), None)
+  private var sourceA: Option[AudioSourceState] = None
   @SuppressWarnings(Array("scalafix:DisableSyntax.var"))
-  private var sourceB: AudioSourceState = new AudioSourceState(BindingKey("none"), None)
+  private var sourceB: Option[AudioSourceState] = None
   @SuppressWarnings(Array("scalafix:DisableSyntax.var"))
-  private var sourceC: AudioSourceState = new AudioSourceState(BindingKey("none"), None)
+  private var sourceC: Option[AudioSourceState] = None
 
-  def playAudio(sceneAudio: SceneAudio): Unit = {
+  def playAudio(sceneAudioOption: Option[SceneAudio]): Unit = {
+    val sceneAudio = sceneAudioOption.getOrElse(SceneAudio.Mute)
+
     updateSource(sceneAudio.sourceA, sourceA).foreach { src =>
       sourceA = src
     }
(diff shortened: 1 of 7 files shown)


davesmith00000:
Looks interesting! I probably won't get around to reviewing it today but I'll try and get around to it soon.

This is one of those areas of the code base that has been used but not exactly battle-tested. Good to get some real world feedback. :smile:

solarized-fox:
thanks! I'm glad my feedback can be helpful! I've tested this with my own game and the audio example and haven't observed any changes in functionality, but of course it's probably worth it for you to look at it too.

for what it's worth, the other reason I'm mucking around in the audio code is because I have to hack in support for playback speed. when even the testers mention "hey shouldn't the audio speed up as the score increases?" I can't exactly say it's out-of-scope for my game 😅. in some regards I wrote this code to double check I understand the AudioPlayer before I attempt to add an actual feature to it. that said, I dunno if _that_ change will end up being a PR too, that'll depend on how good I feel about the solution I come up with.

and as one final aside, I've failed to mention it until now I think, but I have been enjoying indigo, minor bumps aside it's a pretty solid and easy-to-use engine. I'm looking forward to sharing the tiny game I'm working on!

davesmith00000:
Hello! Sorry about the slow reply - I've been on holiday for a week. :tropical_drink: 

Really glad to hear you are enjoying using Indigo and can't wait to see what you're building! I love seeing peoples creations!

Regarding the speed up / slow down of the sound: That is a good idea, I hadn't thought of that. Even if you don't want to submit a PR, I think it would be a great feature that I'm happy to support, and it would be helpful to take a look at your solution sometime/somehow. Let me know if you need any help. :smile:

I shall get on and review this PR.

indigo/indigo/src/main/scala/indigo/platform/audio/AudioPlayer.scala
```diff
@@ -149,32 +151,36 @@ final class AudioPlayer(context: AudioContextProxy) {
...
 
+        next.playbackPattern match {
           case PlaybackPattern.SingleTrackLoop(track) =>
-            currentSource.audioNodes.foreach(_.audioBufferSourceNode.stop(0))
-
             val nodes =
               findAudioDataByName(track.assetName)
-                .map(asset => setupNodes(asset, track.volume * sceneAudioSource.masterVolume, loop = true))
+                .map(asset => setupNodes(asset, track.volume * next.masterVolume, loop = true))
+                .get //throws if no asset found
```
davesmith00000, line 173: Commenting just for posterity: I think this is fine. I have this long running internal debate about what to do if an asset you need just isn't there, and bailing out is the only sensible thing to do at the moment. As a user you can check for the presence of your asset before you use it but it's a slightly awkward process. It's similar territory to having a backend Java service that's been running for weeks and then tries to reflectively load something that doesn't exist - what do you do?

davesmith00000 (approved):
I think this looks great! Thank you for the work and sorry again that it's taken me so long to review it.

When I wrote this code the first time around I was (am?) going through a phase or preferring ADTs to anything else, but I think just using `Option` here makes a lot of sense. :+1:

indigo/indigo/src/main/scala/indigo/platform/audio/AudioPlayer.scala
davesmith00000: Also the `Outcome` type does let you handle stop-the-world issues like this (in theory).

PR #416, opened 2022-09-18
Timeline Animations

davesmith00000:


davesmith00000:
I threw together a mock DSL yesterday to see how this could be presented to the user:

image

...but it has lots of problems. Not least that you can't describe parallel vs sequential operations. Some suggestions/feedback I got back included:

1. Use of `and` and `then` to describe parallel and sequential respectively.
2. Possibility of merging timelines.

To help work through this, I thought I might try looking at the following made up use cases:

1. delay start, fade logo in, pause, fade logo out, pause, emit event. (One property, sequential change.)
2. Same as above, but move from left to right continuously through out (Parallel animation of same entity, different property. Overlapping times)
3. Same as above, but during pause, ease-in-out the scale (Parallel animation during one segment.)

davesmith00000:
animated-crate

Gif quality is a bit poor but here is the latest animation. Appear after 2 seconds (gif may be clipped), move from `Point(0, 0)` to `Point(100, 100)`, then to `Point(100, 0)` using ease-in-out for both journeys.

davesmith00000:
At this point, an animation is a `SignalFunction`, and this willbe expensive for big animations:

```scala
  val modifier: Graphic[Material.Bitmap] => SignalFunction[Seconds, Graphic[Material.Bitmap]] =
    SF.easeInOut(5.seconds) >>> SF.lerp(Point(0), Point(100)) >>> move(_)
```

davesmith00000:
Things left to do:

- [ ] More helper methods
- [ ] Tests
- [ ] Clips
- [ ] Sprites
- [ ] Review naming?
- [ ] Emit events

...then we're done enough for a first edition I think.

PR #439, opened 2022-10-19
Shader lang

davesmith00000:
Very much a work in progress...


davesmith00000:
Where I'm up to:

(There's a [board](https://github.com/PurpleKingdomGames/indigo/projects/9?add_cards_query=is%3Aopen) of some missing stuff I've thought of...)

I have a code generator that manufactures all the swizzle combinations.
I have a macro that given a quite specific program written in Scala, is known to produce GLSL that works, see images below:

fp-shader-maybe-6_scala_glsl

image

Works perfectly, but there will be many combinations of program that don't work yet. Also my macro code is horrible at the moment. Many of the supporting classes are nice and have tests, but the macros look like the gnarly work of a professional cargo-cult'er, and this isn't far from the truth. My next job is to firm that stuff up - or at least clean it up a bit.

davesmith00000:
Redundant and antiquated. We have Ultraviolet now.

PR #444, opened 2022-11-29
Convert Indigo Shaders from GLSL to Ultraviolet

davesmith00000:


davesmith00000:
Ah yeah, the build is going to fail until I publish Ultraviolet.

Chicken and egg, I need to do enough of this work to prove UV is ready for release. Works just fine locally.

davesmith00000:
Ported `ImageEffects`, once again, worked first time...
image

davesmith00000:
Quick update:
There are only three 'library' shaders left to port: `ShapeBox`, `ShapeCircle`, and `ShapePolygon`.

After that, a decision needs to be made about the 'base' template shaders, and if (:crossed_fingers:) they are/can be replaced, then the GLSL lang validator and all the shader code gen can also be removed.

davesmith00000:
Some of the things that are left to do that I know about:

1. There are at least two representations of types like `vec2`, can we just use the ultraviolet ones everywhere?
2. Is it possible to use existing hand-rolled GLSL shaders with the new UV base shaders?
3. Examine the developer experience of using UV shaders and see if it's good enough.
4. Rename WebGL2Base and WebGL2Merge to Entity and Blend respectively.
5. Restructure the base shaders to allow use of named functions - needs a different composition approach.
6. Try and reuse some of the shader library shaders to make, say, a 9-slice shader and test reusability.

davesmith00000:
[ultraviolet_indigo_test.webm](https://user-images.githubusercontent.com/908709/214154718-8c0b4841-bf61-420f-b7ef-9d30ec489422.webm)
Just because I like PR's with pictures.

davesmith00000:
I think this is ready, but I need to cut a release of Ultraviolet to prove it works...

davesmith00000:
The build is green!

...but I've thought of more things to do:

1. The standard shaders are still using the old GLSL bases.
2. The new UV shader bases still contain the lighting stuff... but that isn't used.

Issue #460, opened 2023-02-11
Add a CI check for code formatting


davesmith00000:
This is long overdue... will also require me to format the code base...

PR #464, opened 2023-02-12
Scalafmt checks on build

davesmith00000:
Fixes issue https://github.com/PurpleKingdomGames/indigo/issues/460


davesmith00000:
As you might expect, there's a lot of formatting noise in this one, but I've broken up the commits into formatting and actual config changes.

Issue #456, opened 2023-01-30
Pointer events support


grzegorz-bielski:
Indigo doesn't have support for touch screens. It can be brought through [Pointer Events API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events). This implementation could also cover existing mouse support, making it a facade on top of pointer events. In the future we could also add some helpers for common gestures.

Original discussion: https://discord.com/channels/716435281208672356/716435281208672359/1069683337238691861

Will also resolve https://github.com/PurpleKingdomGames/indigo/issues/55

davesmith00000:
Did you manage to get any further with this @grzegorz-bielski?

No pressure, I only ask because I'm hoping to cut a new release in the next couple of weeks and it would be an awesome addition. If you're not going to be ready though, don't sweat it. :smile:

grzegorz-bielski:
Still on it! You can follow the progress on [this branch](https://github.com/grzegorz-bielski/indigo/commits/pointer-events-support).

What's left is adding some gesture handlers to make the pointers events support more useful from the get go ... and further testing.
Not sure if I'll be able to finish it for the next release though.

Issue #55, opened 2020-09-13
Add Touch events


davesmith00000:
This may be slightly more involved than I'd anticipated.

MDN Reference:
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#Specifications

Spec:
https://www.w3.org/TR/touch-events/#touch-interface

grzegorz-bielski:
What about pointer events?
MDN doc: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events
This is a slightly newer API that covers more input methods and unifies mouse and touch events.

davesmith00000:
Thanks for the tip! I'm generally quite behind the times on the latest browser developments, always good to get some steer. 😄 

Also compatibility looks to be fine: https://caniuse.com/pointer

davesmith00000:
Closing this as it will be superseded by @grzegorz-bielski excellent work adding pointer support.

PR #471, opened 2023-02-18
Pointer events support

grzegorz-bielski:
Resolves: https://github.com/PurpleKingdomGames/indigo/issues/456 and https://github.com/PurpleKingdomGames/indigo/issues/55

Changes:
- Basic pointer events support.
- Fixed a bug with not removed event handlers - `removeEventListener` doesn't remove the hadlers added by `element.onevent = fn`
- Based mouse events support on pointers events
- PointersScene for manual QAing

Yeah, no gestures...

I was planning on adding them, but this turned out to be quite a rabbit hole, and I might not have the time to finish it in the near future. Moreover, I think they need to be first implemented in some game based on Indigo's pointer events, then after some polishing, contributed to Indigo, not the other way around.
My initial idea is that as gesture tracking can be heavy, they should be opt-in, enabled only if some `GestureArea(onDrag=fn)` component is used, similar to existing `HitArea`. There shouldn't be `PointersState` on `InputState`, or it has to be very thin. I could create another ticket for it.


indigo/sandbox/src/main/scala/com/example/sandbox/scenes/PointersScene.scala
```diff
@@ -0,0 +1,100 @@
+package com.example.sandbox.scenes
+
+import com.example.sandbox.Fonts
+import com.example.sandbox.SandboxAssets
+import com.example.sandbox.SandboxGameModel
+import com.example.sandbox.SandboxStartupData
+import com.example.sandbox.SandboxViewModel
+import indigo.*
+import indigo.scenes.*
+import indigo.shared.events.PointerEvent
```
davesmith00000, line 10: Can you add an alias to this to the [package](https://github.com/PurpleKingdomGames/indigo/blob/33a63baefb212fe21f7179783c05b36f930e3e97/indigo/indigo/src/main/scala/indigo/package.scala#L372) so that it just arrives with the standard imports?

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -108,37 +108,29 @@ object MouseButton:
   */
 sealed trait MouseEvent extends InputEvent:
   val position: Point
-  val x: Int
-  val y: Int
+  val x: Int = position.x
+  val y: Int = position.y
```
davesmith00000, line 112: Good idea :+1:

indigo/indigo/src/main/scala/indigo/shared/collections/Batch.scala
```diff
@@ -298,8 +299,10 @@ object Batch:
 
   private[collections] final case class Combine[A](batch1: Batch[A], batch2: Batch[A]) extends Batch[A]:
     val isEmpty: Boolean      = batch1.isEmpty && batch2.isEmpty
-    def head: A               = batch1.head
-    def headOption: Option[A] = batch1.headOption
+
+    export batch1.head
+    export batch1.headOption
+    export batch1.lastOption
```
davesmith00000, line 305: :+1:

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -189,16 +173,89 @@ object MouseEvent:
...
+    */
+  opaque type Buttons = Int
+  object Buttons:
+    inline def apply(buttons: Int): Buttons = buttons
+
+    given CanEqual[Buttons, Buttons] = CanEqual.derived
+
+  /** Pointing device is moved into canvas hit test boundaries. It's counterpart is [[PointerLeave]].
+    */
+  case class PointerEnter(position: Point, pointerId: PointerId, buttons: Buttons, isPrimary: Boolean)
```
davesmith00000, line 228: All case classes should be `final` please. These's a bunch here and also one in the demo scene at the bottom.

davesmith00000:
This a great piece of work - many thanks.

I left a couple of very minor improvement comments.

Can I just check if your work has been updated with the latest work on `main`? I only ask because there was a lot of it, though I don't think in this area.

One other question: Do you think we should mark the mouse events as deprecated?

davesmith00000 (commented):
Great work, few minor comments.

grzegorz-bielski:
> Can I just check if your work has been updated with the latest work on `main`? I only ask because there was a lot of it, though I don't think in this area.

I already rebased it against newest main and resolved conflicts - there were only some minor ones in the sandbox.

> One other question: Do you think we should mark the mouse events as deprecated?

I'm not sure. Probably not yet? There is no APIs around PointerEvents like `indigo.shared.input.Mouse` yet, so it might be confusing.

davesmith00000:
> There is no APIs around PointerEvents like indigo.shared.input.Mouse yet

Good point. So in fact what we're probably missing is a `Pointer` equivalent of `Mouse`, and then we could mark `Mouse` (and mouse events) as deprecated.

But we can have that as a separate issue if you'd prefer, so that it doesn't become a blocker for this work?

grzegorz-bielski:
> So in fact what we're probably missing is a `Pointer` equivalent of `Mouse`, and then we could mark `Mouse` (and mouse events) as deprecated.

Maybe even the `Pointer` could "extend" current `Mouse`, like pointer events are extending mouse events in the DOM API. 
Like mentioned in the og post, I think the pointer specific calculations, like gesture recognizers should be rather put in `indigoextras.ui` with `HitArea` and similar, so the _actual_ `Pointer` state could be quite thin.

> But we can have that as a separate issue if you'd prefer, so that it doesn't become a blocker for this work?

Yes, definitely. I wanted to push something out so people can start experimenting. The merger of current mouse events with pointer events and abstractions on top if it needs more thought.

I believe we can create two new issues: 
- Gesture handlers impl 
- Integrating `Pointer` state with `Mouse` state.

davesmith00000:
Sounds good, would you like to go ahead and create the two follow up issues? The descriptions don't need to be massive, just enough to help the next person (no doubt you or I!) remember what the plan was.

I'll checkout this PR later and give it a test run just to sanity check it - I always do - then hit the merge button. :+1: 

Thanks again!

davesmith00000 (approved):
Works great! :fire: 

...and I'm frankly shocked that you can paint the whole window in circles and still get 120fps. True it's degraded from about 550fps by that point, but still! We could up the performance by switching to clones, but that's not important for this work. :smile:

davesmith00000:
I'm going to go ahead and create those two issues. I'm got time, might as well.

Issue #476, opened 2023-02-18
Integrate `Pointer` state with `Mouse` state.


davesmith00000:
In PR #471, @grzegorz-bielski introduced Pointer events to Indigo, but we do not have an accompanying `InputState` model for it. `MouseEvent` comes with a `Mouse` model, for example.

It would be good to add that and consider replacing or deprecating `Mouse` and `MouseEvent` in the process.

Issue #475, opened 2023-02-18
Add Gesture support


davesmith00000:
Add gesture handling to indigo-extras.

See this PR by @grzegorz-bielski and related issues for details: https://github.com/PurpleKingdomGames/indigo/pull/471

PR #503, opened 2023-03-22
Add basic Pointers state

grzegorz-bielski:
While [tinkering with gestures](https://github.com/grzegorz-bielski/pure-tetrio/blob/gesture-controls/frontend/src/main/scala/pureframes/tetrio/game/scenes/gameplay/viewmodel/TapGestureArea.scala) I discovered that I actually need access to sequence of pointer events from current frame. The easiest way to do it was to modify Indigo by adding barebones Pointers state.

While this doesn't resolve the https://github.com/PurpleKingdomGames/indigo/issues/476 or https://github.com/PurpleKingdomGames/indigo/issues/475 it could be a starting point that makes the experimentation a little bit easier.

I'd be nice to merge it before 0.15.0 😄

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/input/Pointers.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/input/Pointers.scala
@@ -1,0 +1,24 @@
+package indigo.shared.input
+
+import indigo.shared.collections.Batch
+import indigo.shared.datatypes.Point
+import indigo.shared.datatypes.Rectangle
+import indigo.shared.events.PointerEvent
+
+final class Pointers(
+    val pointerEvents: Batch[PointerEvent],
+    val position: Point
+)
+
+object Pointers:
+  val default: Pointers =
+    Pointers(Batch.empty, Point.zero)
+
+  def calculateNext(previous: Pointers, events: Batch[PointerEvent]) =
+    Pointers(
+      pointerEvents = events,
+      position = lastPointerPosition(previous.position, events)
+    )
+
+  private def lastPointerPosition(previous: Point, events: Batch[PointerEvent]): Point =
+    events.collect { case mp: PointerEvent.PointerMove => mp.position }.lastOption.fold(previous)(identity)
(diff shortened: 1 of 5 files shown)


indigo/project/Misc.scala
```diff
@@ -27,7 +27,7 @@ object Misc {
     UsefulTask("", "sandboxRun", "Run the sandbox game (fastOptJS + Electron)"),
     UsefulTask("", "perfRun", "Run the perf game (fastOptJS + Electron)"),
     UsefulTask("", "shaderRun", "Run the shader game (fastOptJS + Electron)"),
-    UsefulTask("", "scalafmtCheckAll", "Launch VSCode"),
+    UsefulTask("", "scalafmtCheckAll", "Check formatting"),
```
davesmith00000, line 30: Thanks! :pray:

indigo/indigo/src/main/scala/indigo/shared/input/Pointers.scala
```diff
@@ -0,0 +1,24 @@
...
+    Pointers(Batch.empty, Point.zero)
+
+  def calculateNext(previous: Pointers, events: Batch[PointerEvent]) =
+    Pointers(
+      pointerEvents = events,
+      position = lastPointerPosition(previous.position, events)
+    )
+
+  private def lastPointerPosition(previous: Point, events: Batch[PointerEvent]): Point =
+    events.collect { case mp: PointerEvent.PointerMove => mp.position }.lastOption.fold(previous)(identity)
```
davesmith00000, line 24: No need to change it, just a comment: I've seen fold used like this a few times recently and I just don't understand how it's syntactically desirable over `getOrElse`. Maybe you can enlighten me on it's appeal?

davesmith00000 (approved):
Looks like a great start for Pointer state! Thanks for the work. :smile: 

With regard to Indigo 0.15.0, I think I'm nearly there. I'm testing it hard as we speak. Hopefully soon...

indigo/indigo/src/main/scala/indigo/shared/input/Pointers.scala
grzegorz-bielski: It was just a straight forward pattern match at first which I directly translated to fold, but yeah the `getOrElse` could have been even better.  I can change it later.

PR #511, opened 2023-05-14
Circle Datatype

davesmith00000:
While chatting to @zetashift, I knocked up this first commit for a new Circle type.

However, I tried to incorporate it into `Shape.Circle` and two things happened:

1. I had a name collision `Shape.Circle` and `Circle` - yuk!
2. I realised that the complementary type to `Rectangle` is not `Circle`, but `Oval`. I may need an `Oval` and a `BoundingOval` type!


davesmith00000:
In the end, I got over the two points above. Oval's will have to wait. :)

Issue #542, opened 2023-06-18
Add Bounding Box/Circle `lineIntersectsInfo`


davesmith00000:
We have super handy `lineIntersects` and `lineIntersectsAt` methods on `BoundingBox` at the moment.

First thing: Do we have that for `BoundingCircle`? ~If not, add them.~ We do, all good.

Next, what I'd like is a `lineIntersectsInfo` method for both, which is like an extended version of `lineIntersectsAt`, providing (assuming impact):

1. The point of intersection
2. The normal of the impact point against the shape
3. The normalised angle of incidence
4. The normalised angle of reflection

PR #546, opened 2023-06-21
Reflection data

davesmith00000:
This PR allows use to shoot a line segment at another line segment, or bounding box, or bounding circle, and get information back about exactly where it struck, the normal, the incident, and the reflected vectors, which we can turn into another LineSegment if we like.


davesmith00000:
[line-reflection.webm](https://github.com/PurpleKingdomGames/indigo/assets/908709/8a2724e2-b294-4917-99d9-a05f8561a82d)

This is it working.

davesmith00000:
Fixes #542

PR #549, opened 2023-07-04
Fixes a minor typo in the documentation

hobnob:
The documentation for `StorageEvent.Loaded` event incorrectly states that the event has a single property (data). However, this event has 2 properties one for `key` and another for `data`


davesmith00000 (approved):
Thanks!

PR #567, opened 2023-07-13
Mill upgrade

hobnob:
This upgrades Mill in Indigo to version 0.11.1

There's also a fix for a warning I was getting while compiling, so might be worth having a gander at that @davesmith00000  and making sure that upgrading it hasn't broken anything in Scala 2 land (I'm fairly sure it hasn't, but an extra pair of eyes is always welcome)


davesmith00000 (approved):
LGTM :+1:

PR #577, opened 2023-07-18
Mouse and Pointer Events

### LINKED ISSUE #512: Add Mouse/Pointer Enter/Leave events
At the moment it's not possible to detect if the mouse or pointer has left the screen. This means if you're listening for a `MouseDown` event we don't know that the mouse has left the game screen (and may subsequently fire a `MouseUp` event we don't capture)

hobnob:
This is a breaking change, and I'm not sure I'm happy about it, so a second (or third) pair of eyes would be very welcome.
This change adds a majority of the properties found in both [Mouse Events](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent) and [Pointer Events](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent). The upshot is that in doing so a simple `Click(position)` match now becomes `Click(position, _, _, _, _, _, _, _)`, however we now have a wealth of new information about the event itself. some key changes are:

## Key Down Properties
Various helper properties (`isAltKeyDown` etc.) to check if a key was held down during the event. The difference here compared to, say, keeping a track of key up/down events within Indigo, is that if the button is held down outside of the scope of the game (such as in another window) this property will still be correct, whereas Indigo will think the button is not held down.

## Buttons Properties
The `buttons` proeprty was previously just on `Pointer` events, but is also available for the `Mouse`. As such, this property has now been moved to the `Mouse` event and updated to be a `Batch` of `MouseButton`

## Mouse Enter/Leave
We already have a Pointer Enter/Leave - we now have the same for the mouse, which fixes #512 

## Tilting, Twisting, and Preassure
Pointer events have been updated with Tilt X/Y, Twist, and Preassure properties. More details on these can be found on the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events). The main change from the docs here is that Twist and Tilt properties are in `Radian`s not degrees.

## Limited Firing of Events
Previously we fired mouse events every time we registered a pointer event. We now only do this if the pointer type is `Mouse`.

### DIFF
--- a/indigo/docs/quickstart/hello-indigo.md
+++ b/indigo/docs/quickstart/hello-indigo.md
@@ -377,7 +377,8 @@ def updateModel(
     context: FrameContext[Unit],
     model: Model
 ): GlobalEvent => Outcome[Model] = {
-  case MouseEvent.Click(clickPoint, _, _, _, _, _, _, _) =>
+  case e: MouseEvent.Click =>
+    val clickPoint       = e.position
     val adjustedPosition = clickPoint - model.center
 
     Outcome(
--- a/indigo/docs/quickstart/hello-indigo.md
+++ b/indigo/docs/quickstart/hello-indigo.md
@@ -86,7 +86,7 @@ import scala.scalajs.js.annotation.JSExportTopLevel
 @JSExportTopLevel("IndigoGame")
 ```
 
-Indigo games are Scala.js projects. We've worked hard to make Indigo feel as much like a normal Scala project as possible, however, we do need a hook for the page. If you're using the standard Indigo Mill or SBT plugins, you ***must name your game "IndigoGame" or it won't work***. Once you move to your own page embed you can call it whatever you like!
+Indigo games are Scala.js projects. We've worked hard to make Indigo feel as much like a normal Scala project as possible, however, we do need a hook for the page. If you're using the standard Indigo Mill or SBT plugins, you _**must name your game "IndigoGame" or it won't work**_. Once you move to your own page embed you can call it whatever you like!
 
 `IndigoSandbox` takes two type parameters that define your start up data type, and the type of your model. Later on we'll introduce a real model, but for now we're just using `Unit` to say "I'm not using these".
 
@@ -263,7 +263,7 @@ graphic.moveBy((d * 600).toInt, 0)
 
 That is, we say that we want a velocity of 600 pixels per second, but multiply that 600 by the fraction of a second since the last frame update. At 60 FPS, `600 * 0.01666 = 9.996` i.e near as makes no odds the 10 pixel movement we wanted, while at a dip to 55 FPS we get `600 * 0.01818 = 10.908`, meaning that you move a little further to make up for lost time.
 
-> This is known as ***frame independent movement***.
+> This is known as _**frame independent movement**_.
 
 ## It isn't a game, if you can't play with it
 
@@ -377,7 +377,7 @@ def updateModel(
     context: FrameContext[Unit],
     model: Model
 ): GlobalEvent => Outcome[Model] = {
-  case MouseEvent.Click(clickPoint) =>
+  case MouseEvent.Click(clickPoint, _, _, _, _, _, _, _) =>
     val adjustedPosition = clickPoint - model.center
 
     Outcome(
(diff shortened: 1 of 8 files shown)


indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
```diff
@@ -111,55 +140,252 @@ final class WorldEvents:
...
+            Radians.fromDegrees(e.twist),
+            pointerType,
+            e.isPrimary,
+            MouseButton.fromOrdinalOpt(e.button)
+          )
         )
-        MouseButton.fromOrdinalOpt(e.button).foreach { button =>
-          globalEventStream.pushGlobalEvent(MouseEvent.MouseDown(position, button))
+
+        if pointerType == PointerType.Mouse then {
```
davesmith00000, line 255: I'm not sure we want that do we? I was under the impression that pointer events were also used for things like touch input?

indigo/docs/quickstart/hello-indigo.md
```diff
@@ -377,7 +377,7 @@ def updateModel(
     context: FrameContext[Unit],
     model: Model
 ): GlobalEvent => Outcome[Model] = {
-  case MouseEvent.Click(clickPoint) =>
+  case MouseEvent.Click(clickPoint, _, _, _, _, _, _, _) =>
```
davesmith00000, line 380: Hmmm yes that is a lot. It's probably the right thing to do but I'm going to sleep on it.

indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
```diff
@@ -194,4 +420,33 @@ final class WorldEvents:
...
+      )
+
+  extension (e: dom.PointerEvent)
+    def width(magnification: Int): Int =
+      (e.width / magnification).toInt
+
+    def height(magnification: Int): Int =
+      (e.height / magnification).toInt
+
+    def indigoPointerType =
```
davesmith00000, line 444: `toPointerType`?

davesmith00000 (commented):
Broadly looks good. Going to have another look with fresh eyes.

I'm mostly concerned about the `pointerType == PointerType.Mouse` filters. Another person was last working on this stuff and I'd been convinced that we should be moving to Pointer events rather than mouse since they were more broadly applicable and allowed us to support additional input types. This looks like we're filtering them out again, but I could be wrong, what's the deal?

indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
hobnob: Yes they are, but we should really only fire things like `MouseDown` if the pointer is actually a mouse. To be honest this is a deviation from the browser, which will fire a `MouseDown` even if what actually occured was a touch event. Happy to keep the more standard browser behaviour of course, I personally feel this is more 'correct', but there's an argument I think for both approaches

hobnob:
> I'm mostly concerned about the `pointerType == PointerType.Mouse` filters. Another person was last working on this stuff and I'd been convinced that we should be moving to Pointer events rather than mouse since they were more broadly applicable and allowed us to support additional input types. This looks like we're filtering them out again, but I could be wrong, what's the deal?

There's definitely an argument to be had here for removing mouse events altogether, especially as `PointerEvent`s inherit from `MouseEvent`s, so you're getting no additional information with regards to the `Mouse`. The difficulty then comes later if you want to deal with `Touch`. `TouchEvent`s (as fired by `touchStart` etc) have additional information about which Touches changed and where they started but don't inherit from `PointerEvent`s. The question then becomes do we just ignore this and continue only supporting `Pointer`, or do we also support `Touch`, and in that case shouldn't we also support `Mouse` for completeness? For now, I think we should carry on supporting the mouse, and if we filter on the pointer type then that means the event itself can be used as a differentiator rather than having to do a match based on pointer type... but I could be wrong, it could be that people would prefer to filter on the pointer type rather than the event itself.

indigo/docs/quickstart/hello-indigo.md
hobnob: I've had a thought about this. I thin (in this instance at least) `case e: MouseEvent.Click` would be a better way of expressing this. I've updated the examples to favour that approach

davesmith00000:
Right, I'm with you now, thanks for the explanation. :+1:

davesmith00000 (approved):
LGTM, thanks @hobnob!

PR #579, opened 2023-07-20
Adds online/offline events

hobnob:
Adds the capability to check whether the application has network access. The caveat here is that having network access does not guarentee internet access, only local network access. Regardless, it should work as a nice indicator for the player to know whether they are experiencing a network issue or not (for example if their router goes down).

### DIFF
--- a/indigo/docs/platform/networking.md
+++ b/indigo/docs/platform/networking.md
@@ -3,7 +3,12 @@ id: networking
 title: Networking
 ---
 
-Indigo supports basic networking via HTTP or WebSockets.
+Indigo supports basic networking via HTTP or WebSockets. You can also get a
+rudimentary indication as to whether the local machine has an active network
+connection using `indigo.platform.networking.Network.isOnline`. Note, however,
+that this will only indicate the presence of a local network and is not
+necessarily an indication of internet connectivity, or the availability of an
+internet resource.
 
 ## Network calls and the game loop
 
--- a/indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
@@ -305,6 +305,18 @@ object KeyboardEvent {
   */
 final case class PlaySound(assetName: AssetName, volume: Volume) extends GlobalEvent
 
+/** A class of events representing general networking events
+  */
+trait NetworkEvent extends GlobalEvent
+
+/** The network has come online and is now available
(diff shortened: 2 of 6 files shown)


indigo/docs/platform/networking.md
```diff
@@ -3,7 +3,12 @@ id: networking
 title: Networking
 ---
 
-Indigo supports basic networking via HTTP or WebSockets.
+Indigo supports basic networking via HTTP or WebSockets. You can also get a
+rudimentary indication as to whether the local machine has an active network
+connection using `indigo.platform.networking.Network.isOnline`. Note, however,
+that this will only indicate the presence of a local network and is not
+necessarily an indication of internet connectivity, or the availability of an
+internet resource.
```
davesmith00000, line 11: I appreciate you updating the docs. I'm also conscious that I have a branch where all the docs have changed. :grimacing:

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -289,6 +289,18 @@ object KeyboardEvent {
...
+  */
+trait NetworkEvent extends GlobalEvent
+
+/** The network has come online and is now available
+  */
+case object Online extends NetworkEvent
+
+/** The network has gone offline and is now unavailable
+  */
+case object Offline extends NetworkEvent
```
davesmith00000, line 302: Rather than looking at things like `NetworkSendEvent`, which is an oddity, can you look at things like `StorageEvent` for the convention here? Specifically it ought to be a `sealed trait`, and the `Online`/`Offline` placed inside a companion object*.

(* This pattern was considered good practice in Scala 2, and Scala 3's enums which are kind of a replacement now enforce it.)

davesmith00000 (dismissed):
Looks good @hobnob - thanks! I've approved the PR as there are no deal breakers here, however I have added a couple of comments, mostly about keeping to Indigo/Scala's general conventions.

davesmith00000:
Ah, there is one thing you're missing! All the events have aliases in the `package.scala` file so there they are discoverable when people do `import indigo.*`. Please add the new ones!

https://github.com/PurpleKingdomGames/indigo/blob/c4d33922194bf88eeb07bddb772b1b71d5e848c8/indigo/indigo/src/main/scala/indigo/package.scala#L403

davesmith00000 (approved):

davesmith00000 (approved):

PR #581, opened 2023-07-20
fix: fixing path finding method for close impassable staying

### LINKED ISSUE #580: Incorrect path locating
Current implementatio of path locating doesn't properly working for some cases. Simple wrong cases are:
```
| | | |   ,   | |?| | 
|?|X|!|       | |X| | 
| | | |       | |!| | 
```
where `?` is start, `!` is end and `X` is impassable cell

This happens because when building edges impassable cells are not ignored, but should
I will attach PR

cheesterx3:
This request contains two changes:

1. Fix for issue #580 
2. Optimizing path building

During local tests optimized version of path building is two times faster on some situations because of filtering a large number of remaining squares which will not be added to edges in any case

Fixes #580

### DIFF
--- a/indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/SearchGrid.scala
+++ b/indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/SearchGrid.scala
@@ -79,13 +79,29 @@ object SearchGrid {
           scored ++ unscored
 
         case (remainingSquares, lastScoredLocations) =>
+          // Calculating bounds coords of checked locations
+          val minMax: (Coords, Coords) = minMaxCoords(lastScoredLocations)
+          val maxCoords = minMax._2
+          val minCoords = minMax._1
+          
+          // Filtering only those coords which are in bounds of checked coords
+          // It will let us to minimize check locations later 
+          val filteredRemaining = remainingSquares.filter(gs => gs.coords.x >= minCoords.x - 1
+            && gs.coords.x <= maxCoords.x + 1
+            && gs.coords.y >= minCoords.y - 1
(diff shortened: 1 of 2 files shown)


davesmith00000:
Thank you @cheesterx3 - I've kicked off a build.

This was one of those: "Who... wrote this.. oh man it was me five years ago" moments. :sweat_smile: Way overdue some love!

davesmith00000:
Just glancing over the code - not your problem just reminding myself how it works: There's a lot of use of `List` and an occasional `foldLeft` in the original implementation - both of which are very slow in Scala.js. If would be interesting to add this to the benchmarks, and then swap `List` for `Batch` (which isn't fun because `Batch`'s pattern matching isn't nice) and remove the `foldLeft`'s and see if the numbers move. :thinking:

cheesterx3:
> Thank you @cheesterx3 - ~I've kicked off a build~ the CI build failed with a formatting error.
> 
> This was one of those: "Who... wrote this.. oh man it was me five years ago" moments. sweat_smile Way overdue some love!

Fixed formatting.

cheesterx3:
> Just glancing over the code - not your problem just reminding myself how it works: There's a lot of use of `List` and an occasional `foldLeft` in the original implementation - both of which are very slow in Scala.js. If would be interesting to add this to the benchmarks, and then swap `List` for `Batch` (which isn't fun because `Batch`'s pattern matching isn't nice) and remove the `foldLeft`'s and see if the numbers move. thinking

Btw, while testing perfomance of path finding `foldLeft` was not an issue - main problem was in filtering large list

davesmith00000 (approved):
Apologies for the slow review. Looks great, thank you!

PR #582, opened 2023-07-24
Storage Event Errors

### LINKED ISSUE #17: Missing "Storage failure" events
We have no events for what happens if we can't read or write to local storage.

hobnob:
Adds error events for `Storage` such that saveing/loading etc can trigger an error which can then be macthed in Indigo.

Fixes #17

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
+++ b/indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
@@ -48,23 +48,52 @@ final class GlobalEventStream(
 
     // Storage
     case StorageEvent.FetchKeyAt(index) =>
-      eventQueue.enqueue(StorageEvent.KeyFoundAt(index, storage.key(index)))
+      storage.key(index) match {
+        case Right(err) => eventQueue.enqueue(err)
+        case Left(str)  => eventQueue.enqueue(StorageEvent.KeyFoundAt(index, str))
+      }
 
     case StorageEvent.FetchKeys(f, t) =>
       val keys = (f to t).toList.map(i => i -> storage.key(i))
-      eventQueue.enqueue(StorageEvent.KeysFound(keys))
+      val errors = keys.flatMap {
+        _ match {
+          case (_, Right(err)) => Some(err)
+          case _               => None
+        }
+      }
+
+      if errors.length > 0 then errors foreach (e => eventQueue.enqueue(e))
+      else
+        eventQueue.enqueue(StorageEvent.KeysFound(keys.flatMap {
+          _ match {
+            case (i, Left(str)) => Some((i, str))
+            case _              => None
+          }
+        }))
 
     case StorageEvent.Save(key, data) =>
-      storage.save(key, data)
+      storage.save(key, data) match {
+        case Right(err) => eventQueue.enqueue(err)
+        case _          => ()
+      }
 
     case StorageEvent.Load(key) =>
-      eventQueue.enqueue(StorageEvent.Loaded(key, storage.load(key)))
+      storage.load(key) match {
+        case Right(err) => eventQueue.enqueue(err)
+        case Left(str)  => eventQueue.enqueue(StorageEvent.Loaded(key, str))
+      }
 
     case StorageEvent.Delete(key) =>
-      storage.delete(key)
+      storage.delete(key) match {
+        case Right(err) => eventQueue.enqueue(err)
+        case _          => ()
+      }
 
     case StorageEvent.DeleteAll =>
-      storage.deleteAll()
+      storage.deleteAll() match {
+        case Right(err) => eventQueue.enqueue(err)
+        case _          => ()
+      }
 
     // Assets
     case AssetEvent.LoadAssetBatch(batch, key, makeAvailable) =>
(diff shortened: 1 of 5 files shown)


indigo/indigo/src/main/scala/indigo/platform/storage/Storage.scala
```diff
@@ -1,23 +1,75 @@
...
+      else Left(Option(dom.window.localStorage.key(index)))
+    catch {
+      case e: js.JavaScriptException =>
+        Right(errToEvent(e, index, StorageActionType.Find))
+    }
 
-  def key(index: Int): Option[String] =
-    Option(dom.window.localStorage.key(index))
+  @SuppressWarnings(Array("scalafix:DisableSyntax.null"))
+  def save(key: String, data: String): Either[Unit, StorageEventError] =
```
davesmith00000, line 25: Not the end of the world, just threw me for a moment. The strong convention is for error's to be on the left, i.e. `Either[StorageEventError, Unit]`. This is because Either is 'right bias' meaning if you do `myEither.map(x => x)`, `x` will be the thing on the right of the either, either your happy path value, rather than the error.

indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
```diff
@@ -48,23 +48,52 @@ final class GlobalEventStream(
...
+        }
+      }
+
+      if errors.length > 0 then errors foreach (e => eventQueue.enqueue(e))
+      else
+        eventQueue.enqueue(StorageEvent.KeysFound(keys.flatMap {
+          _ match {
+            case (i, Left(str)) => Some((i, str))
+            case _              => None
+          }
```
davesmith00000, line 71: It's early, but I think you're using some of Scala's implicit conversions here to automagically turn an `Option` into a `List`. If so I'd prefer:

```scala
            case (i, Left(str)) => List((i, str))
            case _              => Nil
```

I really don't think Scala should let you treat Option as List, it causes much confusion, but here we are!

indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
```diff
@@ -48,23 +48,52 @@ final class GlobalEventStream(
 
     // Storage
     case StorageEvent.FetchKeyAt(index) =>
-      eventQueue.enqueue(StorageEvent.KeyFoundAt(index, storage.key(index)))
+      storage.key(index) match {
+        case Right(err) => eventQueue.enqueue(err)
+        case Left(str)  => eventQueue.enqueue(StorageEvent.KeyFoundAt(index, str))
```
davesmith00000, line 53: Comment further down about Either's being the other way around. In this case I squinted at this for about five minutes convinced there was an error here. There isn't! But it threw me for a while. :smile:

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -299,6 +303,61 @@ trait NetworkReceiveEvent extends GlobalEvent
...
 /** Events relating to manipulating locally stored data
   */
+
+enum StorageActionType:
+  case Save, Load, Delete, Find
+
+sealed trait StorageEventError extends GlobalEventError:
+  /** The identifier of the storage item accessed. Either the index (an Int), or the key (a String)
+    */
+  val key: String | Int
```
davesmith00000, line 313: I'm not sure how I feel about this. There's nothing wrong with it technically, but it feels like we're leaking implementation details / JavaScript.

The alternative would be something like:

```scala
enum StorageKey:
  case Index(value: String), Key(value: String)
```

...but undeniably it amounts to the same thing...

indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
hobnob: That wil give me a `List[List[(Int, String)]]` won't it? Above I'm using `flatMap` to take a `List[(Int, Option[String])]` and convert it to a `List[(Int, String)`. Happy to change it though if I'm wrong :)

indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
```diff
@@ -48,23 +48,52 @@ final class GlobalEventStream(
...
+        _ match {
+          case (_, Left(err)) => Some(err)
+          case _              => None
+        }
+      }
+
+      if errors.length > 0 then errors foreach (e => eventQueue.enqueue(e))
+      else
+        eventQueue.enqueue(StorageEvent.KeysFound(keys.flatMap {
+          _ match {
```
davesmith00000, line 68: No need to fix it now, but in cases like this, this line can be dropped to:

```scala
        eventQueue.enqueue(StorageEvent.KeysFound(keys.flatMap {
          case (i, Right(str)) => Some((i, str))
          case _               => None
        }
```

indigo/indigo/src/main/scala/indigo/platform/events/GlobalEventStream.scala
davesmith00000: We're saying that same thing. I'm saying using flatMap to convert `List[Option[A]]` to `List[A]` is totally allowed by Scala, but I consider it bad practice (because I don't think it should be allowed). Don't worry about it now.

davesmith00000 (approved):
LGTM!

PR #584, opened 2023-07-27
Auto-Resize

### LINKED ISSUE #469: Fullscreen rendering bug (debounce)
When you resize an electron window by dragging the corners around, the image resizes once you let go.

I recall that this delay is handled by the debounce code that was put in many moons ago.

However, when you switch to fullscreen mode you often get a squashed image or an image at the wrong aspect ratio.

Some experimenting tells me that Indigo is getting the correct numbers, and when I was trying this out I was rendering a shader that is quite happy to adapt itself to any aspect ratio, but in this case it ended up looking squashed.

This leads me to suspect that the canvas element itself (or something on the page) is not being resized correctly. I haven't looked, but for example it might be that the debounce code doesn't quite complete in a full screen scenario - or something like that.

The reason we have the debounce code is just to remove the jarring effect of the resize. A simple test would be to disable it. If it is that, we could just move the debounce code into Indigo itself (or make the game builders do the work).

hobnob:
This PR closes #469 by removing `debounce` completely and adding automatic resizing throughout Indigo. This is another breaking change I'm afraid, as the `GameConfig` has now been given a new property called `resizePolicy`, which determines the way Indigo reacts to a resize. The new default behaviour is for Indigo to always resize depending on how big the container is. Whether this is sensible or not remains to be seen, but it felt sensible at the time 🙂 

The resize policy can be changed in the `GameConfig` using either: `withResizePolicy(resizePolicy)`, `noResize`, `autoResize`, or `autoResizePreserveAspect`.

This is all implemented using the [Resize Observer](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver), on the parent container. This is the recomended way of observing size changes, and will fire whenever the containing element changes size and waits for resource to be free before it does so. Indigo will fire an `ApplicationResize` event to let the app know it's being resized (for developer information really), and will then resize the canvas according to the policy, either not taking any action (`NoResize`), resizing to the entire width and height of the container (`AutoResize`) or resizing while maintaining the aspect ratio of the canvas element (`AutoResizePreserveAspect`).

It's worth noting that any CSS that overrides the `width` and `height` attributes of the canvas will interfere with the resizing policy and may produce unexpected results.

### DIFF
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/IndigoBuild.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/IndigoBuild.scala
@@ -32,7 +32,7 @@ object IndigoBuild {
     os.write(directoryStructure.base / "cordova.js", "")
 
     // Write support js script
-    val support = SupportScriptTemplate.template(false)
+    val support = SupportScriptTemplate.template()
     os.write(directoryStructure.base / "scripts" / "indigo-support.js", support)
 
     // Fill out html template
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/IndigoCordova.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/IndigoCordova.scala
@@ -23,7 +23,7 @@ object IndigoCordova {
 
     // Write support js script
     val supportFile = outputDir / "www" / "scripts" / "indigo-support.js"
-    val support     = SupportScriptTemplate.template(true)
+    val support     = SupportScriptTemplate.template()
     os.remove(supportFile)
     os.write(supportFile, support)
 
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/IndigoRun.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/IndigoRun.scala
@@ -36,7 +36,7 @@ object IndigoRun {
 
     // Write support js script
     val supportFile = outputDir / "scripts" / "indigo-support.js"
-    val support     = SupportScriptTemplate.template(true)
+    val support     = SupportScriptTemplate.template()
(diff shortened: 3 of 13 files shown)


davesmith00000 (commented):
Right. All of this looks really good except for two things:

1. Have we lost the delay on resize?

Maybe you're saying we don't need it anymore? But the point of the debounce code was to _not_ resize on every little twitch of the dimensions, because you end up with ugly graphical tearing or squash n' stretch rendering. It also means the WebGL context is being adjusted a lot which is bad news for performance. We've lost the debounce code (yay!) but not replaced it with an alternative, it seems?

2. Do we really need the `ApplicationResize` event?

The real difference between `ApplicationResize` and `ViewportResize` is that the former reports any change in the size of the container whenever that happens, and `ViewportResize` reports any change in the size of the _useable rendering area_ after the rendering context has been reinitialised.

I expect only `ViewportResize` is interesting to the majority users. The only use I can think of for `ApplicationResize`, is manually re-implementing the debounce function as mentioned in point (1).

hobnob:
>     1. Have we lost the delay on resize?

Yes.. but also no. The delay now is down to resource. If JS has enough resource to deal with a resize event then it will, rather than stressing the browser. If it doesn't then it won't and it will wait until it does. This means we're no longer waiting a set time to decide we've waited long enough, the browser is deciding based on the machine capabilities.

>     2. Do we really need the `ApplicationResize` event?

Possibly not. If you don't think it's useful, I'm happy to remove it :)

davesmith00000 (approved):
Excellent work @hobnob - delighted to see the debounce script deprecated! I've tested it locally and it's a big improvement. :tada:

davesmith00000 (approved):

Issue #592, opened 2023-08-28
You can declare multiple asset directories


davesmith00000:

davesmith00000:
Ok, on reflection I've decided not to bother with this. Using the new filters, having specified one based directory, you can organise how you like under that.

Issue #591, opened 2023-08-28
Plugin: Apply filters during asset copy


davesmith00000:
We need to be able to exclude / include files that are in the assets directory.

I think we need both include and exclude options, so that you can generally exclude a folder, but then (for example) take certain parts of it. The rules might be:

1. Take the file if there is a matching include for it, or
2. Ignore the file is there is a matching exclude for it, or
3. Take it.

I wondered if we need some sort of glob matching, but I think we can probably do a lot just with strings and regex's in the first instance.

PR #593, opened 2023-08-28
Plugin config and asset improvements

davesmith00000:
Working on #591 and #592, for generally better asset shifting.


davesmith00000:
From the Mill version, you can now do this:
image

The sbt one is the same, except that you'll use the overload functions that use `String`s instead of `os-lib` paths.

PR #595, opened 2023-09-03
Plugin generators

davesmith00000:
Provide a series of source generators to make working with Indigo and it's assets easier. This PR includes:

- Config: Generates a Scala module with a `GameConfig` instance that is synchronised with your build settings (view port and background colour).
- Asset listing: Generates a Scala moduke listing all your assets, respects folder structure, and provides helpers like sets of assets to load and accompanying simple materials etc.
- Embed text: ...from any text file into a Scala module.
- Embed GLSL shader pairs: Takes a vertex and fragment shader pair and embeds them into a Scala module
- Data: Embed's (for instance) CSV or Markdown tables as either Enum's or Maps
- Aseprite data: Can also be embedded directly into a Scala module


davesmith00000:
I think this is there now, I'm going to test it on the Roguelike to double check it works as expected.

davesmith00000:
This has been tested and extensively expanded / improved through the process of using it on the roguelike in this PR: https://github.com/davesmith00000/roguelike/pull/30

PR #627, opened 2023-11-05
Fixed #612: Friendly mouse event extractors

### LINKED ISSUE #612: Add friendly extractor for MouseEvents
Work was done to expose more browser mouse properties via the `MouseEvent`s - which is great - but at the cost of cumbersome pattern matching.

See if a new extractor can be added just for the position property, which I expect is the common use case, I'd like to be able to:

`case MouseEvent.Click(position) =>  ???`

davesmith00000:
This change makes the bold assumption that 9 times out of 10, you what to know - for example - that the mouse was clicked and where that happened, and that all of the other information is _generally_ less interesting.

So now you can do this:

```scala
case MouseEvent.Click(pt) => ???
```

Instead of:

```scala
case MouseEvent.Click(pt, _, _, _, _, _, _, _) => ???
```

You can still access the other properties, like so:

```scala
case e: MouseEvent.Click => e.button

// Or at the head of a list of events...
case (e: MouseEvent.Click) :: evts => e.button
```

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
@@ -223,6 +223,8 @@ object MouseEvent:
         movementPosition = Point.zero,
         button = MouseButton.LeftMouseButton
       )
+    def unapply(e: Click): Option[Point] =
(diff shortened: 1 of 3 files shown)


bilki (approved):
Nice 👍

PR #629, opened 2023-11-05
Supplying CloneTileData/CloneBatchData as a raw array

davesmith00000:
TODOs:

- [ ] Find out why clone instancing isn't working in Perf game
- [ ] Confetti is slower, I'm thinking it's the flatMap. Refactor code to optimally use raw data. (Or is this just the wrong use case? Test on roguelikes instead?)


davesmith00000:
I've slept on it, and I may park this.

I like the syntax change that happened here, and that might be worth keeping (?), but I think the core idea may be deeply flawed. The issue is that to gain all this efficiency you'd need to basically bind your model structure to the mutable structure this thing requires including all the rendering specific values. If you're going to do that, there's probably much simpler ways to gain efficiency.

davesmith00000:
Parking this PR.

Issue #646, opened 2023-12-02
Pathfinding does not find path


mprevel:
Hi,

I think the pathfinding algorithm may miss to find a valid path.

It can be reproduced with the following snippet:

```
val pathsMap: Array[Array[Char]] = // 12x12
  Array(
    Array('*', '*', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_'), // 0
    Array('*', '*', '_', '_', '_', '_', '_', '_', '_', '_', '_', '_'), // 1
    Array('*', '*', '*', '*', '*', '*', '_', '_', '_', '_', '_', '_'), // 2
    Array('_', '_', '_', '_', '_', '*', '_', '_', '_', '_', '_', '_'), // 3
    Array('_', '_', '_', '_', '_', '*', '_', '_', '_', '_', '_', '_'), // 4
    Array('_', '_', '_', '_', '_', '*', '*', '*', '_', '_', '_', '_'), // 5
    Array('_', '_', '_', '_', '_', '_', '_', '*', '_', '_', '_', '_'), // 6
    Array('_', '_', '_', '_', '_', '_', '*', '*', '_', '_', '_', '_'), // 7
    Array('_', '_', '_', '_', '_', '_', '*', '_', '_', '*', '*', '*'), // 8
    Array('_', '_', '_', '_', '_', '_', '*', '_', '_', '*', '*', '*'), // 9
    Array('_', '_', '_', '_', '_', '_', '*', '*', '*', '*', '*', '*'), // 10
    Array('_', '_', '_', '_', '_', '_', '*', '*', '*', '*', '*', '*') // 11
  )

val n           = pathsMap.length // 12
val startCoords = Coords(0, 0)
val endCoords   = Coords(n - 1, n - 1)
val impassableCoords =
  for {
    x <- 0 until n
    y <- 0 until n
    if pathsMap(y)(x) == '_'
  } yield Coords(x, y)

val searchGrid = SearchGrid.generate(
  start = startCoords,
  end = endCoords,
  impassable = impassableCoords.toList,
  gridWidth = n,
  gridHeight = n
)

```

```
println(searchGrid.isValid) // true
println(searchGrid.locatePath(Dice.fromSeed(0))) // List(Coords(0,0))
```

A possible expectation is:
```
[...]

PR #648, opened 2023-12-03
Add pathfinder

mprevel:
Hi,

This is a pathfinding algorithm as discussed in https://github.com/PurpleKingdomGames/indigo/issues/646

It took me a little bit more time than what I was expecting.

Please let me know if you have questions, comments or want some changes about it.

Regards,

### DIFF
--- a/indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/DefaultPathBuilders.scala
+++ b/indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/DefaultPathBuilders.scala
@@ -11,14 +11,14 @@ object DefaultPathBuilders:
   // defines the movements as relative points
   object Movements:
     // most common movements
-    val Up: Point         = Point(0, -1)
-    val Down: Point       = Point(0, 1)
-    val Right: Point      = Point(1, 0)
-    val Left: Point       = Point(-1, 0)
-    val UpRight: Point    = Up + Right
-    val DownRight: Point  = Down + Right
-    val DownLeft: Point   = Down + Left
-    val UpLeft: Point     = Up + Left
+    val Up: Point        = Point(0, -1)
+    val Down: Point      = Point(0, 1)
+    val Right: Point     = Point(1, 0)
+    val Left: Point      = Point(-1, 0)
+    val UpRight: Point   = Up + Right
+    val DownRight: Point = Down + Right
+    val DownLeft: Point  = Down + Left
+    val UpLeft: Point    = Up + Left
 
     // most common movement groups
     val Vertical: List[Point]   = List(Up, Down)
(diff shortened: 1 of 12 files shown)


indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/DefaultPathBuilders.scala
```diff
@@ -0,0 +1,210 @@
+package indigoextras.pathfinding
+
+import indigo.*
+import indigoextras.pathfinding.DefaultPathBuilders.Movements.*
+
+import scala.scalajs.js
+
+// simple path builder implementations based on Point
+object DefaultPathBuilders:
```
davesmith00000, line 9: Could this all be in `object PathBuilder`? One less term for users to learn?

indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/SearchGrid.scala
```diff
@@ -6,8 +6,11 @@ import indigoextras.pathfinding.GridSquare.EndSquare
 import indigoextras.pathfinding.GridSquare.ImpassableSquare
 import indigoextras.pathfinding.GridSquare.StartSquare
 
+import scala.annotation.nowarn
 import scala.annotation.tailrec
 
+@deprecated("Use the new indigoextras.pathfinding.PathFinder", "0.15.3")
+@nowarn("cat=deprecation")
 final case class SearchGrid(
```
davesmith00000, line 14: It's tempting to just remove this stuff. I doubt many people are using it. :thinking:

davesmith00000 (commented):
Apologies for the slow review. This looks great! :rocket: 

I've left a few minor comments for your consideration - mostly just for style / consistency. I think the next things are:

1. I'd like to see it working / have an example to turn to for reference. Usually at this point I'd set up a scene in the sandbox, maybe just a simple grid and have it solve and show coloured boxes for the path or something. I can do this if you'd rather not.

2. Decide whether or not to just remove the old search grid. I think we could. Indigo isn't in a place where people are moaning about breaking API changes.

We could just raise issues for those two points, and I can make a demo and then consider removing SearchGrid as steps outside this PR.

indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/PathFinder.scala
```diff
@@ -0,0 +1,127 @@
...
+  def heuristic(
+      t1: T,
+      t2: T
+  ): Int // heuristic allows to select the way to estimate the distance from a point to the end
+
+// A* (A Star) inspired algorithm version allowing generic types and customisation of the path finding
+object PathFinder:
+
+  private type OpenList[T]   = List[PathProps[T]] // the open list is the list of the points to explore
+  private type ClosedList[T] = List[PathProps[T]] // the closed list is the list of the points already explored
```
davesmith00000, line 24: Avoid `List` where possible, its very slow compared to `Batch`. :smile:

indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/PathFinder.scala
```diff
@@ -0,0 +1,127 @@
...
+    * @param h
+    *   the heuristic (distance from end)
+    * @param f
+    *   the g + h
+    * @param parent
+    *   the parent point
+    * @tparam T
+    *   the type of the points
+    */
+  private case class PathProps[T](value: T, g: Int, h: Int, f: Int, parent: Option[PathProps[T]] = None)
```
davesmith00000, line 40: Avoid default args, I'd rather be explicit everywhere, almost especially on internal plumbing.

indigo/indigo-extras/src/main/scala/indigoextras/pathfinding/SearchGrid.scala
mprevel: I did not remove it.

mprevel:
The PR is updated according to the comments but the removal of the old pathfinder.

davesmith00000 (approved):
Fantastic work @mprevel, and the demo is very cool. :sunglasses: 
image

Issue #536, opened 2023-06-05
BUG: Drawing a triangle using `Shape.Polygon` doesn't seem to work brilliantly


davesmith00000:
![image](https://github.com/PurpleKingdomGames/indigo/assets/908709/e8979f1e-49c3-4496-af87-43b46c124c59)

Points to mull over:

- We can't currently specify whether the Polygon is open or closed and there are no docs to say which it is.
- There's a weird line at the top, I wonder if that line has always been there and I just haven't noticed before (based on my limited test cases)?
- Is it worth adding a `Shape.Triangle` for more accurate drawing support?

JPonte:
I have the same issue and it's not just triangles.
I'm drawing some prallelograms and it works perfectly on indigo 0.14.0 but that line started appearing when I updated to 0.15.0.

davesmith00000:
Ah! That's a good clue though! Indigo 0.14.0 still used the old GLSL shaders, seems likely that I've just messed up the port of the polygon shader from GLSL to Scala. Thanks! :pray:

davesmith00000:
Fixed by @JPonte :rocket: 
https://github.com/PurpleKingdomGames/indigo/pull/654

PR #654, opened 2023-12-12
Fix Polygon shader

JPonte:
#536 

The i and j vars are being initialized outside of the loop and then immediatly at the start of every loop, which means that the initial values are actually never used.

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/shader/library/ShapePolygon.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/shader/library/ShapePolygon.scala
@@ -52,14 +52,14 @@ object ShapePolygon:
         var j: Int = count - 1
 
         while i < count do
-          j = i
-          i = i + 1
           val e = v(j) - v(i)
           val w = p - v(i)
           val b = w - e * clamp(dot(w, e) / dot(e, e), 0.0f, 1.0f)
           d = min(d, dot(b, b))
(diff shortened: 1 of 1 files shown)


davesmith00000:
Any chance of a screenshot? :smile:

indigo/indigo/src/main/scala/indigo/shared/shader/library/ShapePolygon.scala
```diff
@@ -52,14 +52,14 @@ object ShapePolygon:
...
-          j = i
-          i = i + 1
           val e = v(j) - v(i)
           val w = p - v(i)
           val b = w - e * clamp(dot(w, e) / dot(e, e), 0.0f, 1.0f)
           d = min(d, dot(b, b))
           val c = bvec3(p.y >= v(i).y, p.y < v(j).y, e.x * w.y > e.y * w.x)
           if (all(c) || all(not(c))) s = s * -1.0f
+          j = i
+          i = i + 1
```
davesmith00000, line 62: Yep that makes sense. I can see why I did that but it's also clearly wrong in hindsight. :facepalm: :smile:

JPonte:
Using this batch of Points:
```scala
val vertices = Batch(
  Point(32, 0),
  Point(32, 32),
  Point(0, 32)
)
```

Before:
Screenshot 2023-12-12 221825

After:
Screenshot 2023-12-12 221741

davesmith00000 (approved):
Great work! Many thanks! :pray:

Issue #665, opened 2023-12-22
Physics: Look into improving scaling performance


davesmith00000:
Indigo's prototype physics engine works ok for what it is (there's a few wrinkles to iron out and lots of features missing, see other open issues) but one issue that needs addressing is that it doesn't scale well.

At the moment, in the best case the engine is probably n^2 performance in all cases.

I'm completely fine with that at the moment for a first release. If you need to simulate a hundred things it'll barely affect your frame rate, and the implementation is pretty clean and easy to follow (and therefore, easy to improve). If however you need to simulate 1000 things or more, then it really starts to hurt.

Having read around the subject, essentially when I've done is implement the "narrow band" phase (where everything checks everything else) as the complete solution. You can't avoid the narrow band phase, it's fine, but what we need is a "broad band" phase ahead of it to sort / cull the scene into small groups of things that need to be tested for collisions, at which point we switch back to narrow band.

It's always a trade off. For a small number of things, adding the broad phase is pure overhead. But when you have many things that overhead might save a lot of work, and hopefully get us towards n log n performance. Might be worth allowing people to configure the broad phase a bit, maybe.

***Note:*** I did try adding caching to the process as a random idea, however this made performance worse since the cost of the caching outweighed the cost of simply doing the calculations.

So I'm thinking we need a tree of some kind, probably my old favourite the `QuadTree`, to organise the colliders spatially (broad phase) before we do the collision resolution.

davesmith00000:
Done, will be in the next release. I'm conscious that the tunnelling problem is the next big thing to sort out.

Issue #576, opened 2023-07-16
Physics: Solve the 'bullet through paper' problem


davesmith00000:

davesmith00000:
Notes to self:

On the physics-noodling branch, the engine is now fast enough to show that the tunnelling problem will always exist with the current implementation (seems obvious in hindsight) because fundamentally, we have to assume 60fps as an upper iteration bound, and if the object is moving sufficiently quickly, it will eventually skip over something.

So what to do?

Currently the engine is a sort of crude 'discrete' engine, in that it really just compares complete moves per iteration. That works perfectly ok (or ok enough) for anything doing a "safe" move, by which I'm meaning (possibly wrongly) that the next position is overlapping the previous position because it's moving slowly enough. In that scenario, the likelihood of tunnelling _seems_ acceptably small (and so edge cases seem ignore-able, at least for now), and so I can treat these as the happy path and just use the existing approach for the calculations.

For "unsafe" items, what I need to do, is do "some" iterations from the old to new positions - ideally leaving no gaps but maybe there's an upper bound to attempts? - and estimate a TOI (time of impact) when it hits each thing (including other unsafe items, so that means iterating them together...), and then solve the collision(s) with the earliest TOI(s), and skip anything else. For example, if at the end of it's move (TOI: 1), a ball hits another ball, but half way there it would have hit a wall (TOI: 0.5), then the ball to ball collision never happened, and we just need to do the ball to wall collision.

PR #666, opened 2023-12-24
Physics engine: Improved scaling performance

davesmith00000:
Casually noodling around with the Physics engine. The main goal is just to make it scale better, but I made try and look at some other problems along the way.

Relates primarily to: https://github.com/PurpleKingdomGames/indigo/issues/665


davesmith00000:
I've started improving the QuadTree implementation so that I can use it here... I've done some stuff already but here is the TODO list:

    Need to implement:
    - Get rid of the 'replace' notion during insert
    - Get rid of remove element 'at' and replace with remove by search.
    - Max depth: The maximum number of sub-divisions allowed.
    - Min size: The smallest allowed quad size before we give up and group all remaining results here.
    - Multple values: Quad's can hold a Batch of values of the given type
    - Max values: Quad's can hold a max value before sub-division unless max depth or min size have been hit.
    - Detect duplicates. If a split results in quads that do not change the outcome, stop and group, to prevent infinite depth due to matching values.
    - Can we switch to Quad.Leaf syntax?
    - Move out of extras to Indigo proper
    - SpatialOps instance for Circle + tests
    - SpatialOps instance for BoundingCircle + tests
    - SpatialOps instance for Line + tests
    - SpatialOps instance for LineSegment + tests
    - Check benchmarks.

davesmith00000:
Running some benchmarks. The changes made so far do not appear to have meaningfully affected performance, but the new implementation has many benefits and is more accurate, so I'm calling that a win.

Before
quadtree_benchmark_before

After
quadtree_benchmark_after

davesmith00000:
With the QuadTree updated, I'm back to looking at the Physics. Quickly adding in the QuadTree as our broad band culling phase has yielded promising results. 1000 circles used to run at about ~25 fps, and now it's up to 60 fps for a while, until the GC (speculation) kicks in and it drops to 45-50 fps. Still, a good improvement for minor code change.

image

Things TODO:

- [x] Add the QuadTree as a broad phase cull
- [ ] Think about how to supply `QuadTree.InsertOptions`, currently I'm making an educated guess. Good enough?
- [ ] Consider optimising the collision code. Currently I do many, many loops, some of which are duplicates.
- [ ] Solve the bullet through a paper wall problem.
- [ ] Consider a mutable solution... does that mean a mutable QuadTree?

davesmith00000:
I was going to have a go at the tunnelling problem (https://github.com/PurpleKingdomGames/indigo/issues/576) here too, but I think I'm going to wrap up this PR now and come back to that in a separate piece of work. There's a lot here already.

PR #682, opened 2024-01-10
Mouse button filter was incorrect

### LINKED ISSUE #681: List of pressed mouse buttons is inverted in mouse events
Hi,

I think I found a bug. I was wondering why my event handling logic did not hook in at

```scala
    case e: MouseEvent.Move if e.buttons.contains(LeftMouseButton) =>
      ...
```

I found the method `indigoButtons` which maps the browser mouse event model into the Indigo event model. I think its condition is wrong:

https://github.com/PurpleKingdomGames/indigo/blob/e1f600a481b940c0334666caa48e72aa14dd3941/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala#L541

Should not it be:

```scala
 .filter(i => ((e.buttons >> i) & 1) == 1)
```

?

hobnob:
Fixes #681

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
+++ b/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
@@ -538,7 +538,7 @@ final class WorldEvents:
     def indigoButtons =
       Batch.fromArray(
         (0 to 5)
-          .filter(i => ((e.buttons >> i) & 1) == 0)
+          .filter(i => ((e.buttons >> i) & 1) == 1)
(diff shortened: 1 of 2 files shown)


davesmith00000:
Following our discussion @hobnob, I've added a unit test for this. I've done it the dumbest way imagine-able, but at least we can be confident it works. :smile: 

Note that due the nature of bitwise ops, the code works with any Int. I've only tested the acceptable / expected range the browser will provide, but it works just fine outside that (effectively a modulo).

(cc @stefanfromm)

davesmith00000 (approved):

PR #686, opened 2024-02-05
Ensures fonts are auto-generated correctly

hobnob:
This ensures that font files are loaded as `Font` asset types and also gives direct access to the `FontFamily` for that font. Supported fonts are `ttf`, `woff`, and `woff2`

### DIFF
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
@@ -262,6 +262,7 @@ object AssetListing {
 
   val FontFileExtensions: Set[String] =
     Set(
+      "eot",
       "ttf",
       "woff",
       "woff2"
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
@@ -163,7 +163,7 @@ object AssetListing {
             val safeName = toSafeName(name, ext)
 
             val vals =
-              s"""${indentSpacesNext}val ${safeName}: AssetName               = AssetName("${name}")
+              s"""${indentSpacesNext}val ${safeName}: AssetName               = AssetName("${name}_${ext}")
               |${indentSpacesNext}val ${safeName}FontFamily: FontFamily    = FontFamily(${safeName}.toString())"""
 
             val loadable =
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
@@ -159,6 +159,21 @@ object AssetListing {
 
             (vals, loadable, named)
 
+          case PathTree.File(name, ext, path) if FontFileExtensions.contains(ext) =>
+            val safeName = toSafeName(name, ext)
+
+            val vals =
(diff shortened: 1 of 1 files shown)


indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
```diff
@@ -245,4 +260,10 @@ object AssetListing {
       "tiff"
     )
 
+  val FontFileExtensions: Set[String] =
+    Set(
+      "ttf",
+      "woff",
+      "woff2"
```
davesmith00000, line 267: `eot` is another one. Arguably we can ignore it, but then, the image list includes `tiff` files so I think the pattern is to do the official thing and leave the user to sort it out.

indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
```diff
@@ -159,6 +159,21 @@ object AssetListing {
 
             (vals, loadable, named)
 
+          case PathTree.File(name, ext, path) if FontFileExtensions.contains(ext) =>
+            val safeName = toSafeName(name, ext)
+
+            val vals =
+              s"""${indentSpacesNext}val ${safeName}: AssetName               = AssetName("${name}")
+              |${indentSpacesNext}val ${safeName}FontFamily: FontFamily    = FontFamily(${safeName}.toString())"""
```
davesmith00000, line 167: I'm running out of time today - just digging around this in the background as you know.

Superficially this all looks good - and adding the `FontFamily` instance is a good idea! - the catch is that I'm worried there are subtle bugs here due to the flaky nature of font loading. There might not be, I'm just unsure and feeling paranoid.

The cause of the problem you correctly identified is the `.` in the file name, and sure enough, the way we register the font is:

```scala
  def loadFontAsset(fontAsset: AssetType.Font): Future[LoadedFontAsset] =
    IndigoLogger.info(s"[Font] Loading ${fontAsset.path}")

    val font = new FontFace(fontAsset.name.toString, s"url(${fontAsset.path.toString})")

    font.load().toFuture.map { fontFace =>
      IndigoLogger.info(s"[Font] Success ${fontAsset.path}")

      // add font to document
      js.Dynamic.global.document.fonts.add(font)
      // enable font with CSS class
      js.Dynamic.global.document.body.classList.add("indigo-fonts-loaded")

      LoadedFontAsset(AssetName(fontFace.family))
    }
```

So the dot would be used, even though it's invalid. But the `ext` in the asset name was helping keep the assets unique, and I'm not sure the use of `safeName` on the next line will necessarily always line up to the same name you've register the font against.

I need to think about this a little more.

davesmith00000 (commented):
Couple of comments, I don't have answers just yet. The solution is nearly there, I'm just worried about how all the naming lines up.

indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
hobnob: Yup, agree :)

indigo-plugin/indigo-plugin/src/indigoplugin/generators/AssetListing.scala
hobnob: Yes, that makes sense. OK, well seeing as it's just the dot that's causing issues, I'll just change the dot to an underscore and all should be fine

davesmith00000 (approved):
Thanks @hobnob, sorry I've been slow to sort this out. I've tested it and it appears to work just fine. :+1:

PR #696, opened 2024-02-22
Updates the Mouse Wheel to allow for more than 1 axis

hobnob:
The mouse wheel event previously had an `amount` property, which represented the movement on the Y axis of a mouse wheel (the up/down motion). However, many devices support more than 1 axis (such as side scrolling), and so this is now represented by the properties `deltaX`, `deltaY`, and `deltaZ`.

The 2 parameter constructor and deconstructor remain the same (assigning the `amount` to the `deltaY`), and so the tests for this remain the same.

This is, however, a breaking change, with anyone who currently uses the `amount` property needing to change that accessor to `deltaY`.


davesmith00000 (approved):
Great work @hobnob! :+1:

PR #708, opened 2024-03-10
Fixed #707: Xbox One Controller not working on FF/Chrome

### LINKED ISSUE #707: Examples crashing with Xbox One S controller
Possible bug when using an Xbox One S controller in example games like The Cursed Pirate, Snake or Roguelike Tutorials. Basically pressing any button makes the whole game crash, and if we check the console we can see they are not recognized by `GamepadInputCaptureImpl`. In my case I was using the controller connected to a USB port, but I imagine this could happen too using the BT connection. I'm going to check both Indigo code and the examples code to find out what is the problem.

imagen

bilki:
Basically a few tweaks here and there to get it working on FF and Chrome (on Linux).

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/platform/input/GamepadInputCaptureImpl.scala
+++ b/indigo/indigo/src/main/scala/indigo/platform/input/GamepadInputCaptureImpl.scala
@@ -20,36 +20,49 @@ object GamepadInputCaptureImpl {
         GamepadInputCaptureImpl.giveGamepadState
     }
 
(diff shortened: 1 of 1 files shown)


davesmith00000 (approved):
All looks perfectly sensible to me (given the current state of gamepad support), also excellent detective work in the issue. :detective: 

I'm happy with it as-is, and will merge it if there's no other feedback, but I'm also going to see if I can get @hobnob to take an interest because he just _loves_ stuff like this. :grin:

davesmith00000:
Have discussed with @hobnob and he's happy to. :smile:

PR #759, opened 2024-08-17
Add vector convenience functions.

yilinwei:
Add 'crossProduct' and alias 'invert' to 'unary_-'.


davesmith00000 (approved):
Looks good, thanks!
If you feel like adding unary_- to Vector4, that would nice too as I'm bound to forget!

davesmith00000 (approved):
LGTM!

Issue #750, opened 2024-07-13
`rollRange` under certain circumstances very likely to return the same result


felher:
The default java.util.Random handles calls to `nextInt(x)` where `x` is a power of two specially.

In Indigo, this leads to the fact that when you do `dice.rollRange(1, 4)` at the start of every `FrameTick`, you are overwhelmingly likely to get the same number as before.

Possible workaround: call `rollRange(1, 4)`  (or with any other number) and discard the first return value. Successive calls seem to yield more random values.

davesmith00000:
@felher, @hobnob has a PR open against this, if you're interested. I don't believe he managed to replicate the issue:

https://github.com/PurpleKingdomGames/indigo/pull/770

I wonder if it's a difference in random behaviour due to the difference in platform? Or was something missed?

JD557:
> I wonder if it's a difference in random behaviour due to the difference in platform? Or was something missed?

The problem might be fixed (I didn't test it) but I think you did miss something,

The issue is not on the distribution of random values (e.g. calling `dice.rollRange(1, 4)` multiple times) but on the distribution of the first random value (e.g. calling `Dice.diceSidesN(4, scala.util.Random.nextLong()).rollRange(1, 4)` multiple times)

felher:
@davesmith00000 @hobnob 
wow, cool, thanks for your work!

I'm not completely sure whether that fixes the issue. @JD557 is right that consecutive random numbers from the same generator should work, but it seems that every frame creates the RNG anew, and the first random number from a dice with 4 sides seems to be highly correlated to the next first random number.

How should I best test the fix? Checkout the repo, publishLocal and test against that? Or is there a snapshot published?
I did dig around a bit and found an old test of mine: https://scribble.ninja/u/felher/cmbyvskqdvghhlgeecbuzibjpug

The first line tells you the number of times the first random number for a frame tick was equal to first number of the last frame tick. The second line tells how many times the second number was equal to the first in the same frame tick. The second number is at 25%, which seems correct given a 4-sided dice. The first number is 90%, though.

hobnob:
> How should I best test the fix? Checkout the repo, publishLocal and test against that?

Yup, that's how you would test this fix :slightly_smiling_face: 

[...]

PR #770, opened 2024-10-07
Adds unit tests to assure randomness

hobnob:
Tests now run 3 lots of 200,000,000 rolls of dice and check the distribution of the results in order to test the assertion made in #750

There's a minor code cleanup here too, but nothing breaking (just re-using a method rather than having 2 similar methods)

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/dice/Dice.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/dice/Dice.scala
@@ -168,9 +168,7 @@ object Dice:
 
       val r: Random = new Random(seed)
 
-      def roll: Int =
-        r.nextInt(sanitise(sides)) + 1
-
+      def roll: Int = roll(sides)
       def roll(sides: Int): Int =
         r.nextInt(sanitise(sides)) + 1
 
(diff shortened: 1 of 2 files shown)


davesmith00000:
There's a lot of good stuff in here that I don't want to lose, however there are two points to have another think about:

1. The discussion on the issue has continued and needs review.
2. I'm not sure about the Date change made here.

Dice are supposed to be 'predicatably unpredicatable'. By salting the seed with the initial date, I feel like we lose a bit of that. Now, I don't know in practice if it makes a great deal of difference: Is anyone really testing that way? Seems unlikely, but I'm also not sure if adding the date will improve the issue being reported. If not then it's an arbitrary change we don't need.

davesmith00000:
Branch needs a rebase.

hobnob:
Done :)

indigo/indigo/src/main/scala/indigo/gameengine/GameLoop.scala
```diff
@@ -21,8 +21,11 @@ import indigo.shared.time.Millis
 import indigo.shared.time.Seconds
 
 import scala.collection.mutable
+import scala.scalajs.js.Date
 import scala.scalajs.js.JSConverters._
 
+val initialDate = new Date()
```
davesmith00000, line 27: Please remove / move it inside the class.

davesmith00000 (commented):
One change to make then I'm happy to merge. Thanks.

Issue #735, opened 2024-05-01
Game screenshots for save game slots


davesmith00000:
It's common practice to take a screenshot of your game and use it as a visual representation of a save game.

Two problems:

1. Saving a screenshot (generally useful?)
2. Loading and using the image data

PR #771, opened 2024-10-23
Screen Capturing

hobnob:
Allows for capturing a screen using the context, fulfilling #735

Usage:
```scala
val screenshot: Asset.Image =
  context
      .captureScreen(
          ScreenCaptureConfig
              .default
              .withName("screenshot1")
              .withExcludeLayers(Batch(uiKey))
   )
```
**Breaking Change:**
Also adds an additional property to `AssetEvent.AssetBatchLoaded` detailing the assets that were loaded in the last `AssetEvent.LoadAssetBatch` event


indigo/indigo/src/main/scala/indigo/platform/renderer/ScreenCaptureConfig.scala
```diff
@@ -0,0 +1,194 @@
...
+    imageType: ImageType
+) {
+
+  /** Set the name of the capture
+    *
+    * @param name
+    * @return
+    */
+  def withName(name: String): ScreenCaptureConfig =
+    this.copy(name = Some(name))
```
davesmith00000, line 36: I noticed you doing this in the Sandbox part below and though "ah no big deal" but this is a core type so I'd better mention it:

Generally you want to avoid using the `Some(..)` constructor unless you're damn sure you're not going to get a `null` value. `Some` gives no `null` safety protection:

```scala
Some(null) // Some(null)
Option(null) // None
```

davesmith00000: On the other hand, using `Some` directly is marginally faster, so in performance sensitive code (which this isn't) where confidence levels are high, you might want to use `Some` instead of `Option`.

indigo/indigo/src/main/scala/indigo/shared/FrameContext.scala
```diff
@@ -29,10 +30,12 @@ final class FrameContext[StartUpData](
     val dice: Dice,
     val inputState: InputState,
     val boundaryLocator: BoundaryLocator,
-    _startUpData: => StartUpData
+    _startUpData: => StartUpData,
+    _renderer: => Renderer
```
davesmith00000, line 34: My first reaction of "WOW that's scary 😱", after taking a moment, I'm not sure it's as scary as it looks, but this does still feel like a sledge hammer way to provide access to the screen capture methods.

An alternative could be:

```scala
final class FrameContext[StartUpData](
  ...
  _startUpData: => StartUpData,
  _captureScreen: Batch[ScreenCaptureConfig] => Batch[Either[String, AssetType.Image]]
```

And then you expose the two variants here on the framecontext an remove one of them from Renderer, i.e.

`Renderer` only has:
```scala
  def captureScreen(captureConfig: Batch[ScreenCaptureConfig]): Batch[Either[String, AssetType.Image]]
```

And `FrameContext` has both:

```scala
  def captureScreen(captureConfig: Batch[ScreenCaptureConfig]): Batch[Either[String, AssetType.Image]]

  def captureScreen(captureConfig: ScreenCaptureConfig): Either[String, AssetType.Image] =
    captureScreen(Batch(captureConfig)).headOption match {
      case Some(v) => v
      case None    => Left("Could not capture image")
    }
```

indigo/indigo/src/main/scala/indigo/shared/platform/SceneProcessor.scala
```diff
@@ -116,17 +117,27 @@ final class SceneProcessor(
...
+          l.toBatch.map(
+            (
+              l match {
+                case LayerEntry.Tagged(tag, _) => Some(tag)
+                case _                         => None
+              },
+              _
+            )
+          )
+        )
```
davesmith00000, line 130: Is it not desirable to do this after the visible check?

indigo/indigo/src/main/scala/indigo/platform/renderer/webgl1/RendererWebGL1.scala
```diff
@@ -1,16 +1,26 @@
 package indigo.platform.renderer.webgl1
 
+import indigo.AssetName
+import indigo.AssetPath
 import indigo.platform.assets.AtlasId
```
davesmith00000, line 5: I'd be tempted to drop this from the WebGL1 implementation. WebGL1 is very limited - really only good enough to display a message saying "don't use this" - and I'd hope no-one is trying to build anything screen capture worthy in it. What do you think?

indigo/indigo/src/main/scala/indigo/platform/renderer/webgl2/RendererWebGL2.scala
```diff
@@ -222,6 +240,72 @@ final class RendererWebGL2(
...
+          canvas.remove()
+
+          Right(
+            AssetType.Image(
+              AssetName(option.name.getOrElse(s"capture-${System.currentTimeMillis()}")),
+              AssetPath(dataUrl)
+            )
+          )
+        )
+    }
```
davesmith00000, line 307: This looks good. So if I'm reading correctly, the process is:

1. Take the previous scene (so that you can filter out the layers people don't want to capture).
2. Set up a new canvas.
3. Re-render which effectively overwrites the usual canvas context.
4. Grab it, and stuff it into the new canvas.
5. Save it.

Have I got that right?

davesmith00000 (commented):
Looks good. I've left some comments. I think the main one is about passing the renderer to the framecontext.

I think the solution you have here is probably as good a job as you can do without a deeper (and trickier!) integration with the renderer.

The alternative I'm thinking of wouldn't need you to retain the previous data because you'd render to two canvas's at once (if some flag was set), dropping the layers you wanted to skip. Don't know if it would work of course, because I haven't tried it!

indigo/indigo/src/main/scala/indigo/platform/renderer/webgl1/RendererWebGL1.scala
hobnob: I think all the time we have a WebGL1 implementation we should support it.. It doesn't add much work in this case, so I think we should keep it

indigo/indigo/src/main/scala/indigo/shared/platform/SceneProcessor.scala
hobnob: We can't because at the pointer we're at `scene.layers` we have a `LayerEntry`, which doesn't contain the `visible` check.. this is only available after we convert the `LayerEntry` to `Content` in the `toBatch` method of the `LayerEntry`.
We could, of course, filter the `toBatch` method... so I might do that

indigo/indigo/src/main/scala/indigo/shared/FrameContext.scala
hobnob: Yes you're right, that's a better route here - I've updated it

davesmith00000 (approved):
Looks good, thanks @hobnob. 🔥

Issue #766, opened 2024-09-19
Switch from `keyCode` to `code` in keyboard


sherpal:
The keyboard and keyboard events rely on the `keyCode` from the JavaScript event, but this is deprecated [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode).

It is supposed to replace by [`code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) (which is a `String`) representing the key that was pressed/released.

Issue #746, opened 2024-06-13
Add Key.isNumeric / Key.asNumber


davesmith00000:
Would be handly to be able to say "was the key a number" and "optionally give me the key's number, if it was a number. Currently doing it was something like this:

```scala
def keyToNumber(keyCode: Int): Option[Int] =
      if keyCode >= Key.KEY_0.code && keyCode <= Key.KEY_9.code then
        Option(keyCode - Key.KEY_0.code)
      else None
```

But in the real version they could just be properties of the enum or something.

PR #772, opened 2024-10-25
Modifies Keyboard events to use the `code` property

hobnob:
**This is a breaking change**

Fixes https://github.com/PurpleKingdomGames/indigo/issues/766 by adding a new `KeyCode` enum, which is populated with the values from the `code` property of the KeyboardEvent. Additionally, the `Key` class also has a new `location` property, which maps to the `location` property of the KeyboardEvent.

All `Key.*_ARROW` constants are now `Key.ARROW_*` to bring it in line with the key codes that they relate to:
 * `Key.UP_ARROW` is now `Key.ARROW_UP`
 * `Key.DOWN_ARROW` is now `Key.ARROW_DOWN`
 * `Key.LEFT_ARROW` is now `Key.ARROW_LEFT`
 * `Key.RIGHT_ARROW` is now `Key.ARROW_RIGHT`

The KeyboardEvent itself now also supports some new additional properties:
 * `isRepeat` - a boolean indicating if the key is being held down
 * `isCtrlKeyDown` - a boolean indicating if the Ctrl key is pressed
 * `isMetaKeyDown` - a boolean indicating if the Meta key is pressed
 * `isAltKeyDown` - a boolean indicating if the Alt key is pressed
 * `isShiftKeyDown` - a boolean indicating if the Shift key is pressed

Additionally adds `isNumeric` and `asNumeric` methods to `Key` in order to fulfill #746

### DIFF
--- a/indigo/indigo-extras/src/main/scala/indigoextras/ui/InputField.scala
+++ b/indigo/indigo-extras/src/main/scala/indigoextras/ui/InputField.scala
@@ -188,10 +188,10 @@ final case class InputField(
           val next = acc.delete
           rec(ks, next, true, acc.key.map(key => InputFieldChange(key, next.text)))
 
-        case Key.LEFT_ARROW :: ks =>
+        case Key.ARROW_LEFT :: ks =>
           rec(ks, acc.cursorLeft, true, changeEvent)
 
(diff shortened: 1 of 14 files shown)


davesmith00000 (approved):
Looks great! 🎉

Issue #749, opened 2024-07-10
The Pointers class is missing convenience methods


davesmith00000:
...the Mouse version inspects events so that you can ask 'was mouse clicked?' for example.

Issue #747, opened 2024-07-10
Button's do not support Pointer events


davesmith00000:

davesmith00000:
I put this together just to show a workaround, but it may give a clue to a real solution too, perhaps.

https://github.com/PurpleKingdomGames/indigo/commit/f3eba3cf9294c1fd097237738c0eab365b73ff11

Issue #703, opened 2024-02-28
Have a way of discerning click events


hobnob:
At the moment click events only fire on a left mouse click, with no way to discern a right click from any other click event. We should fire the click event when a right **or** left click happen, allow discerning which button was clicked, and allow any pointer device clicks.

davesmith00000:
I've had a look into this, two things:

1. There is no Right click, but we could latch onto the [contextmenu event](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event), what do you think?
2. There is a workaround. You can listen for pointer events, like this:

```scala
    case c @ PointerEvent.PointerUp(_) =>
      println(c.button) // Tells you which button was released, e.g. : `RightMouseButton`
      Outcome(viewModel)
```

Issue #692, opened 2024-02-14
Pointer Events: FrameContext.Pointers looses pointer positions


hobnob:
`FrameContext.pointers` is meant to keep track of the current position/tilt/buttons etc of all of the pointers being used. However, currently as soon as the main pointer (in my scenario a mouse) stops moving the pointer is removed from the `Batch` of pointers. I suspect this is because we treat the pointers in a similar way to the mouse events, but it needs a bit more thinking about and tinkering with

davesmith00000:
I've had a quick look and the implementations are very different. I think `Mouse` could do with a bit of a clean up, and `Pointers` probably needs to work a bit more like mouse does.

It would be helpful to list a couple of examples of the expected / desired behaviours. (I don't really use Pointers :sweat_smile:)

PR #774, opened 2024-10-28
Fixes a number of Pointer related issues

### LINKED ISSUE #748: SceneContext does not export the pointer instance
As it does with 'mouse'.

hobnob:
Addresses the following issues:
* Adds `pointers` to `SceneContext` (#748)
* Add new methods to `Pointers` ( #749)
* Changes the way events are captured for pointers ( #692)
* Adds a new `PointerEvent.PointerClick` method, which is fired when a `PointerDown` event is followed by a `PointerUp` event for the same button within a specified, configurable, time frame (defaults to 200ms) (#703)
* Updates UI to use `Pointer` rather than `Mouse` (#747)

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
+++ b/indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
@@ -24,6 +24,9 @@ import org.scalajs.dom
 import org.scalajs.dom.document
 import org.scalajs.dom.html
 import org.scalajs.dom.window
+import scala.scalajs.js.Date
+
+var pointerButtons: Map[Double, Batch[(Int, Date)]] = Map.empty
 
 final class WorldEvents:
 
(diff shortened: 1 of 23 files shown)


indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
```diff
@@ -23,6 +23,11 @@ import org.scalajs.dom.document
 import org.scalajs.dom.html
 import org.scalajs.dom.window
 
+import scala.scalajs.js.Date
+
+@SuppressWarnings(Array("scalafix:DisableSyntax.var"))
+var pointerButtons: Map[Double, Batch[(Int, Date)]] = Map.empty
```
davesmith00000, line 29: This can't be here as a top level this, needs to move inside the class at the very least, i.e.:

```scala
final class WorldEvents:
  @SuppressWarnings(Array("scalafix:DisableSyntax.var"))
  var pointerButtons: Map[Double, Batch[(Int, Date)]] = Map.empty
```

davesmith00000: And it should probably be `private`.

davesmith00000 (commented):
This looks good. One big no-no jumped out at me, will check it over again with fresh eyes.

indigo/indigo/src/main/scala/indigo/platform/events/WorldEvents.scala
hobnob: All done

Issue #744, opened 2024-05-27
GameViewport type improvements


davesmith00000:
Perhaps GameViewport can just be an opaque type equal to Size, and export (using Scala 3 export) all of `Size`'s methods?

At the moment, you can't do GameViewport(..) / 2, but you can with a Size.

davesmith00000:
Done

PR #778, opened 2024-10-30
feat: deprecate GameViewport.scala

david-lebl-adastra:
#744 issue

### DIFF
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/generators/ConfigGen.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/generators/ConfigGen.scala
@@ -27,7 +27,7 @@ object ConfigGen {
       |
       |  val config: GameConfig =
       |    GameConfig(
-      |      viewport = GameViewport(${indigoOptions.metadata.width}, ${indigoOptions.metadata.height}),
+      |      viewport = Size(${indigoOptions.metadata.width}, ${indigoOptions.metadata.height}),
       |      frameRateLimit = Option(FPS.`60`),
       |      clearColor = ${extractBgColor(indigoOptions.metadata.backgroundColor)},
       |      magnification = 1,
--- a/indigo/indigo/src/main/scala/indigo/IndigoShader.scala
+++ b/indigo/indigo/src/main/scala/indigo/IndigoShader.scala
@@ -120,7 +120,7 @@ trait IndigoShader extends GameLauncher[IndigoShaderModel, IndigoShaderModel, Un
       model: IndigoShaderModel
   ): GlobalEvent => Outcome[IndigoShaderModel] = {
     case ViewportResize(vp) =>
-      Outcome(model.copy(viewport = vp.size))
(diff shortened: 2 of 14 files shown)


david-lebl-adastra:
Possibly many breaking changes within this commit create during hackathon night. Possible solution with less refactoring using opaque type with context bound, which will allow "export" of Size class methods: 
`opaque type GameViewport <: Size = Size`

davesmith00000:
Hi @david-lebl-adastra!

Thanks for having another go! But actually... I preferred the deprecated version. 🙂

An `opaque type A <: B = B` is the same as `type A = B`, i.e. **it is no longer an opaque type**, it's just a type alias, which is why the export works. `GameViewport` _is_ a `Size` wearing a disguise.

I find type aliases to be useful only in very limited cases ad I tend to avoid them because they're misleading, and hence I'm fairly allergic to the `opaque type A <: B = A` pattern.

The beauty of an opaque type (without the context bounds) is that you can't mistake it for something else at compile time because the compiler won't let you. With the context bounds, you lose all mechanical compiler assistance.

So if the opaque type is just a type alias, then I'd rather not have it at all and embrace the breaking change.

Does that make sense?

hobnob:
Is there a case here for a type alias though? I'm thinking of situations where maybe your method takes a viewport and you want to be clear that's what you're expecting, rather than just any old size?

davesmith00000:
> Is there a case here for a type alias though? I'm thinking of situations where maybe your method takes a viewport and you want to be clear that's what you're expecting, rather than just any old size?

Right but a type alias doesn't help you there. You either want an opaque type (which is a type alias where you _cannot_ replace it with any old `Size`), or your declare bankruptcy and ditch GameViewport in favour of Size, as we originally did here.

Unfortunately the only way to replace GameViewport with a Size as an opaque type, is to manually expose all the methods. Doable, but the question is do we care? (<--- This is probably the better solution! However...)

Generally I've found having GameViewport to be different from Size to just be a hassle. The only nice thing about it is that it has constructors for the various common screen resolutions.

hobnob:
I'm a +1 for an opaque type here then... We don't have any specific need yet for differences between Size and GameViewport, but it's nice to have that option (an Islands Cape method for example), but I'm happy to be outvoted 🙂

davesmith00000:
Thanks again @david-lebl-adastra, sorry about the back and forth. It isn't always easy to know what to do until you try. 😅 

If it's ok I'll push another commit here to bring it inline with the discussion - unless you want to do it?

I think the work is to keep what you've done, but remove the type bounds, so that means:

1. Bringing back the methods that take `GameViewport` types, e.g. `def topLeft(viewport: GameViewport): Point`.
2. Remove the `<: Size`.
3. Fix anything that doesn't compile (might be nothing, I think you've done 90% if not all of the work already, in fact)

What say you?

davesmith00000 (approved):

davesmith00000:
Hmmm we need to bring the fork up to date in order to merge it. I'll see if I can sort it out later.

Issue #783, opened 2024-11-08
Introduce 'services'


davesmith00000:
Recent work around the randomness of `Dice` and the shiny new screen capture tooling, has shone a light on an uncomfortable truth I've been ignoring: `FrameContext` is a bit broken.

The functions on the interface of Indigo are supposed to be pure and referentially transparent, to aid testing. How practical is testing at the top level? I don't know, but that's the idea.

Mostly that's fine, no problems here:

`def update(model: Model): GlobalEvent => Outcome[Model]`

But we're missing `frameContext`:

`def update(frameContext: FrameContext, model: Model): GlobalEvent => Outcome[Model]`

Many of the fields are fine, such as the fields providing the current time and the input states (I think), but others - particularly the boundary locator - are stateful, side effecting, and impossible/hard to stub for testing.

The proposal is to create a notion of 'services' perhaps still available via the frame context somehow, which capture these useful facilities in a way that doesn't break the principles of the interface.

davesmith00000:
Done, will be in the next release.

Issue #742, opened 2024-05-19
BoundaryLocator interface for testing


davesmith00000:
Currently you can't produce a fake framecontext (of any type) because you cannot make a BoundaryLocator. If BoundaryLocator became an interface then the side effects can happen in the concrete instance and fake versions can be stubbed out for testing.

davesmith00000:
Done, will be in the next release.

PR #784, opened 2024-11-09
Replace FrameContext with Context

davesmith00000:
Goals:

1. Introduce the idea of services, which should be stub-able to replace the parts of `FrameContext` that were stateful or side effecting.
2. Rework `FrameContext` into `Context`, something better organised / designed.
3. Also split random number generation into `Dice` (pseudo random and imperfect, but stateless) and `Random` (stateful).

Related issues: https://github.com/PurpleKingdomGames/indigo/issues/783, https://github.com/PurpleKingdomGames/indigo/issues/742, and to some extent https://github.com/PurpleKingdomGames/indigo/issues/750


davesmith00000:
Brief summary:

`FrameContext` replaced with `Context`. `Context` is split into `frame` and `services`, which are frame values and long running processes respectively, accessible something like this:

```scala
// `StartUpData` is accessible at the top level
ctx.startUpData

// Example 'frame' values / facilities
ctx.frame.time.delta
ctx.frame.dice.rollDouble // PRNG seeded on the game running time at the point of frame entry

// Example 'services'
ctx.services.screen.capture(..)
ctx.services.random.nextInt // Long running scala.util.Random instance.
```

PR #824, opened 2025-01-30
Added BlankEntity size modifiers

### LINKED ISSUE #813: BlankEntity is missing resize modifiers
At the moment you can only set the size during construction, usually there would be (I think, check) `withSize`, `resizeTo`, and `resizeBy` modifiers.

exoego:
Closes #813 

Not sure to add `size` override to `transform(To|By)` so skipped.

```scala
  def transformTo(newPosition: Point, newRotation: Radians, newScale: Vector2, newSize: Size): BlankEntity =
    this.copy(position = newPosition, rotation = newRotation, scale = newScale, size = newSize)

  def transformBy(positionDiff: Point, rotationDiff: Radians, scaleDiff: Vector2, sizeDiff: Size): BlankEntity =
    transformTo(position + positionDiff, rotation + rotationDiff, scale * scaleDiff, size = size * sizeDiff)
```

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/scenegraph/BlankEntity.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/scenegraph/BlankEntity.scala
@@ -85,7 +85,7 @@ final case class BlankEntity(
     resizeTo(newSize)
 
   def resizeBy(amount: Size): BlankEntity =
-    this.copy(size = size * amount)
+    this.copy(size = size + amount)
   def resizeBy(w: Int, h: Int): BlankEntity =
     resizeBy(Size(w, h))
 
--- a/indigo/indigo/src/main/scala/indigo/shared/scenegraph/BlankEntity.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/scenegraph/BlankEntity.scala
@@ -77,6 +77,18 @@ final case class BlankEntity(
   def withRef(x: Int, y: Int): BlankEntity =
     withRef(Point(x, y))
 
+  def resizeTo(newSize: Size): BlankEntity =
+    this.copy(size = newSize)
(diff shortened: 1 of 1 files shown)


davesmith00000 (dismissed):
Perfect, thanks. 👍

indigo/indigo/src/main/scala/indigo/shared/scenegraph/BlankEntity.scala
```diff
@@ -77,6 +77,18 @@ final case class BlankEntity(
...
 
+  def resizeTo(newSize: Size): BlankEntity =
+    this.copy(size = newSize)
+  def resizeTo(w: Int, h: Int): BlankEntity =
+    resizeTo(Size(width, height))
+  def withSize(newSize: Size): BlankEntity =
+    resizeTo(newSize)
+
+  def resizeBy(amount: Size): BlankEntity =
+    this.copy(size = size * amount)
```
davesmith00000, line 88: Ah, sorry, just noticed this. `resizeBy` is addition of a relative amount, so `*` should be `+`.

exoego: Ok so only scale used *

exoego, line 88:
```suggestion
    this.copy(size = size + amount)
```

davesmith00000: Correct, yes. The use-case for `resizeBy` is "However big it is, I need it 10x10 pixels bigger".

davesmith00000 (approved):
Thanks! 😄

PR #831, opened 2025-02-02
Stabilize Batch#hashCode

### LINKED ISSUE #740: Does Batch produce a stable hashCode?
If not, add a `hashCode` method.

exoego:
Closes #740

`Batch#hashCode` is now stable regardless of concrete classes (`Wrapped` or `Combine`), since now calculated from elements.
Calculation time is increasing proportionally to number of elements.


davesmith00000 (approved):
Brilliant! Thanks! 🙏

PR #834, opened 2025-02-04
Fix various windowisms

### LINKED ISSUE #833: Building on windows
Various windowisms that make building on windows fail. (Tests fail.)

In particular CRLF where it doesn't matter makes a bunch of tests fail.

When using assertEquals this makes test fail because of spurious CRLF instead of NEWLINE. Switching to assertNoDiff allows these to pass. This also gets rid of the need for associated trims to make expected and actual line-up.

EmbedGLSLShaderPair.extractShaderCode presumes the input string is formatted with leading NEWLINE. When a file is read from windows it is CRLF aka RETURN & NEWLINE. So this function fails.

mtrigueira:
Fixes #833 

Amends tests that fail because of incidental CRLF causing assertEquals to fail to instead use assertNoDiff. This also removes the need for some trims on expected and actual.

Another test fix is to convert to string expressions before comparison, where the CRLF are embedded.

One test retrieves its input from the os, and when constructing the expected value presumes lines are separated by NEWLINE, the regex technique explained next is used to coerce the string into the expected format.

EmbedGLSLShaderPair.extractShaderCode presumes the input string is formatted with leading NEWLINE. When a file is read from windows it is CRLF aka RETURN & NEWLINE. This fix sidesteps the issue by regex replacing all instances of the magical regex "\\R"  newline with exactly NEWLINE.

### DIFF
--- a/indigo-plugin/indigo-plugin/src/indigoplugin/generators/EmbedGLSLShaderPair.scala
+++ b/indigo-plugin/indigo-plugin/src/indigoplugin/generators/EmbedGLSLShaderPair.scala
@@ -97,8 +97,8 @@ object EmbedGLSLShaderPair {
     """.stripMargin
 
   def extractShaderCode(text: String, tag: String, newName: String): List[ShaderSnippet] =
-    s"""//<indigo-$tag>\n((.|\n|\r)*)//</indigo-$tag>""".r
-      .findAllIn(text)
+    s"""//<indigo-$tag>\\n((.|\\n|\\r)*)//</indigo-$tag>""".r
+      .findAllIn(text.replaceAll("\\R", "\n"))
       .toList
       .map(_.toString)
       .map(_.split('\n').drop(1).dropRight(1).mkString("\n"))
--- a/indigo-plugin/indigo-plugin/test/src/indigoplugin/generators/AssetListingTests.scala
+++ b/indigo-plugin/indigo-plugin/test/src/indigoplugin/generators/AssetListingTests.scala
@@ -144,10 +144,8 @@ class AssetListingTests extends munit.FunSuite {
       Set(
         a
       )
-      """.trim
+      """
 
-    assertEquals(actual.trim, expected.trim)
-
+    assertNoDiff(actual, expected)
   }
-
 }
--- a/indigo/indigo/src/test/scala/indigo/shared/shader/library/BaseBlendShaderTests.scala
+++ b/indigo/indigo/src/test/scala/indigo/shared/shader/library/BaseBlendShaderTests.scala
@@ -17,19 +17,25 @@ class BaseBlendShaderTests extends munit.FunSuite {
       }
 
     val actual =
-      BlendShader.vertex(modifyVertex, IndigoUV.VertexEnv.reference).toOutput.code
+      BlendShader
+        .vertex(modifyVertex, IndigoUV.VertexEnv.reference)
+        .toOutput
+        .code
+        .stripMargin
+        .trim
+        .replaceAll("\\R", "\n")
 
     val expected1: String =
       """
       |vec4 vertex(in vec4 v){
       |  return v+vec4(1.0);
       |}
-      |""".stripMargin.trim
+      |""".stripMargin.trim.replaceAll("\\R", "\n")
 
     val expected2: String =
       """
       |VERTEX=vertex(VERTEX);
-      |""".stripMargin.trim
+      |""".stripMargin.trim.replaceAll("\\R", "\n")
 
     assert(clue(actual).contains(clue(expected1)))
     assert(clue(actual).contains(clue(expected2)))
@@ -59,19 +65,23 @@ class BaseBlendShaderTests extends munit.FunSuite {
       }
 
     val actual =
-      BlendShader.fragment(modifyColor, IndigoUV.BlendFragmentEnv.reference).toOutput.code
+      BlendShader
+        .fragment(modifyColor, IndigoUV.BlendFragmentEnv.reference)
+        .toOutput
+        .code
+        .replaceAll("\\R", "\n")
 
     val expected1: String =
       """
       |vec4 fragment(in vec4 v){
       |  return v+vec4(1.0);
       |}
-      |""".stripMargin.trim
+      |""".stripMargin.trim.replaceAll("\\R", "\n")
 
     val expected2: String =
       """
       |COLOR=fragment(COLOR);
-      |""".stripMargin.trim
+      |""".stripMargin.trim.replaceAll("\\R", "\n")
 
     assert(clue(actual).contains(clue(expected1)))
     assert(clue(actual).contains(clue(expected2)))
--- a/build.ps1
+++ b/build.ps1
@@ -1,15 +1,16 @@
-Set-Location -Path indigo-plugin -PassThru
-& .\build.ps1
-Set-Location -Path .. -PassThru
+function build {
+    param (
+        [string]$ProjectPath
+    )
 
-Set-Location -Path sbt-indigo -PassThru
-& .\build.ps1
-Set-Location -Path .. -PassThru
+    echo ">>> $ProjectPath"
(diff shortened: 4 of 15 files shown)


mtrigueira:
Scalafmt wanted a space?

Interestingly when run from powershell the build does not stop on fail. (no "set -e" in the powershell script.

```terminal
[5/5] ================== indigo-plugin[2.12].checkFormat ================= 4s
1 tasks failed
indigo-plugin[2.12].checkFormat Found 1 misformatted files
[5/5] indigo-plugin[2.13].checkFormat                                                                                                                                                    
[5] Checking format of 34 Scala sources                                                                                                                                                  
[5] parsed config (v3.8.5): C:\Users\...\IdeaProjects\indigo\indigo-plugin\.scalafmt.conf
```

I've tweaked this on a branch. Will need to open a separate ticket for that, or at least a separate pull request. It's muddying the waters around this branch. This branch won't be complete until that is resolved first.

Putting the fail fast throws up a new back of problems of a similar nature.

mtrigueira:
Please first merge #835

indigo-plugin/indigo-plugin/test/src/indigoplugin/generators/AssetListingTests.scala
```diff
@@ -144,10 +144,8 @@ class AssetListingTests extends munit.FunSuite {
       Set(
         a
       )
-      """.trim
-
-    assertEquals(actual.trim, expected.trim)
+      """
 
+    assertNoDiff(actual, expected)
```
davesmith00000, line 149: Good to know!

indigo-plugin/indigo-plugin/src/indigoplugin/generators/EmbedGLSLShaderPair.scala
```diff
@@ -97,8 +97,8 @@ object EmbedGLSLShaderPair {
     """.stripMargin
 
   def extractShaderCode(text: String, tag: String, newName: String): List[ShaderSnippet] =
-    s"""//<indigo-$tag>\n((.|\n|\r)*)//</indigo-$tag>""".r
-      .findAllIn(text)
+    s"""//<indigo-$tag>\\n((.|\\n|\\r)*)//</indigo-$tag>""".r
```
davesmith00000, line 100: We'll need to test this somehow. I don't think `\\n` is correct. Or if it is correct, then why does it work at the moment (on linux / mac)?

I think the `OriginalScene` in the sandbox could be run on Mac / Linux and Windows to check it's working. Should look like this (noting the sea in the background):

image

indigo/indigo/src/test/scala/indigo/shared/shader/library/BaseBlendShaderTests.scala
```diff
@@ -17,19 +17,25 @@ class BaseBlendShaderTests extends munit.FunSuite {
...
       |  return v+vec4(1.0);
       |}
-      |""".stripMargin.trim
+      |""".stripMargin.trim.replaceAll("\\R", "\n")
 
     val expected2: String =
       """
       |VERTEX=vertex(VERTEX);
-      |""".stripMargin.trim
+      |""".stripMargin.trim.replaceAll("\\R", "\n")
```
davesmith00000, line 38: In the cases like this, will `assertNoDiff` not save you from having to do the `replaceAll`?

davesmith00000 (commented):
Thanks! A few minor comments to look at.

indigo-plugin/indigo-plugin/src/indigoplugin/generators/EmbedGLSLShaderPair.scala
mtrigueira: The original works without the double escape as the single escape means the "\n" is converted by string and is handed as 0xD to the regular expression. With the double escape "\\n" string emits a "\n". This is then handed to the regular expression which interprets the "\n" as 0xD. By the time the regex performs the match they are equivalent. The convention is to let strings do string escaping, and regex to do regex escaping. It has the added benefit of when you dump the string it is legible as the "\n" is readable as opposed to a rendered 0xD which is whitespace.

The second issue is the windowism. For context here are the lines
```scala
    s"""//<indigo-$tag>\\n((.|\\n|\\r)*)//</indigo-$tag>""".r
      .findAllIn(text.replaceAll("\\R", "\n"))
```
I really didn't like the replaceAll jammed in there, but I didn't want to mess with the regex meaning as expressed. On second thoughts I've tweaked the original regex and either regexString2 or 3 below should work on both *nix and windows.

Once we've decided on what do with this, I'll update the def and put together a unit test capturing the essence of
works on *nix and windows below.

I propose regexString3. It prints nicely, and works on windows.
 
```scala
// scratch.sc
val tag : String = "tag"

val regexString1 : String = s"""//<indigo-$tag>\n((.|\n|\r)*)//</indigo-$tag>"""
val regexString2 : String = s"//<indigo-$tag>\r?\n((.|\r|\n)*)//</indigo-$tag>"
val regexString3 : String = s"//<indigo-$tag>\\r?\\n((.|\\r|\\n)*)//</indigo-$tag>"

// >>> Why double escape?
println("1="+regexString1)
/*
1=//<indigo-tag>
((.|
|)*)//</indigo-tag>
*/
println("2="+regexString2)
/*
2=//<indigo-tag>?
((.|
|)*)//</indigo-tag>
*/
println("3="+regexString3)
/*
3=//<indigo-tag>\r?\n((.|\r|\n)*)//</indigo-tag>
*/

def extractCode(regexString: String, text: String): List[String] =
  regexString.r
    .findAllIn(text)
    .map(_.split('\n').drop(1).dropRight(1).mkString("\n"))
    .toList

// >>> Works on unix.
val nix = "//<indigo-tag>\nOne\n//</indigo-tag>"
println("nix="+nix.replaceAll("\n","<LF>").replaceAll("\r","<CR>"))
//nix=//<indigo-tag><LF>One<LF>//</indigo-tag>
println("1="+extractCode(regexString1, nix))
//1=List(One)
println("2="+extractCode(regexString2, nix))
//2=List(One)
println("3="+extractCode(regexString3, nix))
//3=List(One)

// >>> Original does not work on windows
val win = nix.replaceAll("\n", "\r\n")
println("win="+win.replaceAll("\n","<LF>").replaceAll("\r","<CR>"))
//win=//<indigo-tag><CR><LF>One<CR><LF>//</indigo-tag>
println("1="+extractCode(regexString1, win))
//1=List()
println("2="+extractCode(regexString2, win))
//2=List(One)
println("3="+extractCode(regexString3, win))
//3=List(One)
```

indigo/indigo/src/test/scala/indigo/shared/shader/library/BaseBlendShaderTests.scala
mtrigueira: I've tried to use assertNoDiff wherever possible. 

In these cases the string gets fed into clue and then contains is called. The assert is an assertTrue, so we cant use assertNoDiff.
```scala
    assert(clue(actual).contains(clue(expected1)))
```

davesmith00000:
Merged https://github.com/PurpleKingdomGames/indigo/pull/835. If you can update this branch I'll try and go through the comments a little later. Apologies for the delay.

mtrigueira:
Not satisfied with where these LF's were coming from. I checked all over the place, and found the CRLF in the scala files themselves. This did not compute, I checked I had a local .gitattributes and set it appropriately and did some git-fu and everything gitlike was happy telling me the files were the same. But they still had CRLF. Turns out when you clone the repo, the CRLF is introduced, and it's a nuisance to banish.  

In the end I deleted the repo from my machine 
```shell
git config --global core.autocrlf input
```
And re-cloned.

It all works now. I'll crosspost into #833

davesmith00000:
Ah! Brilliant. Glad you figured it out, and I've learned about `assertNoDiff` as an added bonus. 😄

PR #835, opened 2025-02-05
Fail build fast in windows

mtrigueira:
On windows the build does not abort, and merrily runs to completion.

This changes the main build script to fail fast, the indigo-plugin and mill-indigo build scripts too. These were giving false negatives.

Some boilerplate code put into local functions.

The "meat" of the powershell and bash scripts now look much more alike.

This is a practical dependency for #833

### DIFF
--- a/indigo-plugin/build.ps1
+++ b/indigo-plugin/build.ps1
@@ -1,5 +1,5 @@
 function RunMill {
-    .\millw.bat -i @Args
+    .\millw.bat @Args
     if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
 }
 


indigo-plugin/build.ps1
```diff
@@ -1,14 +1,22 @@
...
+mill clean indigo-plugin[2.12]
+mill clean indigo-plugin[2.13]
+mill indigo-plugin[2.12].compile
+mill indigo-plugin[2.13].compile
+mill indigo-plugin[2.12].test
+mill indigo-plugin[2.13].test
+mill indigo-plugin[2.12].checkFormat
+mill indigo-plugin[2.13].checkFormat
+mill indigo-plugin[2.12].publishLocal
+mill indigo-plugin[2.13].publishLocal
```
davesmith00000, line 22: I nearly left a different comment here, because `mill` is a confusing choice of function name. Can you go for something like `runMill` (or something... get creative). I probably won't look at this script again for months and I don't want to make the same mistake twice. 😄

davesmith00000 (commented):
Looks good (my powershell knowledge is approaching zero, but it looks plausible!), just one small amend if you could.

indigo-plugin/build.ps1
mtrigueira: I've tidied up the function too, and called it "RunMill". I've also moved the "-i" back into the calls because although I was aiming for a nice proxy function, so my brain did not have to deal with the boilerplate. Slurping up the parameter also made me do a double take on second reading. Hopefully this does the trick.

davesmith00000 (approved):
Thank you!

PR #852, opened 2025-03-01
Fixed #843: Allow Indigo to be elegantly halted

### LINKED ISSUE #843: Allow Tyrian to elegantly clean up / remove Indigo games that use the bridge
At the moment it isn't totally clear how to do that (there is a `halt()` method) and @mprevel reports null pointer exceptions.

davesmith00000:
Relates to https://github.com/PurpleKingdomGames/indigo/issues/843

Below is a demo of the tyrian / indigo sandbox halting and removing two games. You can see that no null pointer exceptions are now thrown in the logs, and after a waiting a moment, the memory usage plunges once the GC kicks in.

Essentially all I've done is removed all the nulling out of internal service instances, which means that halting now clears some memory by purging all the caches, but the instance is still present until the SPA (in this case a Tyrian app, but whatever) removes the instance and it's canvas element. After that, all remaining memory is cleaned up.

https://github.com/user-attachments/assets/15b317ce-1d78-43a1-894e-9ca6460a501d

(cc @mprevel)

### DIFF
--- a/indigo/tyrian-sandbox/src/main/scala/example/IndigoSandbox.scala
+++ b/indigo/tyrian-sandbox/src/main/scala/example/IndigoSandbox.scala
@@ -28,6 +28,33 @@ object IndigoSandbox extends TyrianIOApp[Msg, Model]:
     case Msg.NoOp =>
       (model, Cmd.None)
 
+    case Msg.RegisterGames(game1, game2) =>
+      (model.copy(game1 = Some(game1), game2 = Some(game2)), Cmd.None)
+
+    case Msg.HaltGame1 =>
+      model.game1.foreach(_.halt())
+      (model.copy(game1 = None), Cmd.None)
+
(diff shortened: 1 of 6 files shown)


indigo/tyrian-sandbox/src/main/scala/example/IndigoSandbox.scala
```diff
@@ -182,8 +231,12 @@ final case class Model(
     bridge: TyrianIndigoBridge[IO, String, Unit],
     field: String,
     components: List[Counter.Model],
-    randomNumber: Int
+    randomNumber: Int,
+    ensureGame1Div: Boolean,
+    ensureGame2Div: Boolean,
+    game1: Option[MyAwesomeGame],
+    game2: Option[MyAwesomeGame]
```
davesmith00000, line 238: This is probably _not_ the most elegant modelling that could be achieved here. 😁

Issue #804, opened 2025-01-08
Waypoints: Add functions for piecewise linear interpolation


JPonte:
The existing interpolation functions are between two values of `Double`, `Vector2` or `Point`, but for some cases it's useful to interpolate between a larger group of values, which apparently is called [piecewise linear interpolation](https://en.wikipedia.org/wiki/Linear_interpolation#Interpolation_of_a_data_set). Non linear versions exist but this one should be fairly simple to implement.

A generic signature would look something like this:
```scala
case class PiecewiseLerpResult(value: Vector2, derivative: Vector2)

// at should be a value between 0.0 and 1.0
def piecewiseLerp(vectors: List[Vector2], at: Double): PiecewiseLerpResult
```

The two values it returns are the interpolated value and it's derivative which for a use case where the Vector2 represents position, the value would be the current position and the derivative its direction and speed.

And to integrate with Indigo's signals it would look like this:
```scala
// Over a time period
def piecewiseLerp(vectors: Vector2, over: Seconds): SignalFunction[Seconds, PiecewiseLerpResult]

// From 0.0 to 1.0
def piecewiseLerp(vectors: Vector2): SignalFunction[Double, PiecewiseLerpResult]
```

I created a type for the result because I find tuples like `(Vector2, Vector2)` can get confusing. 

An assumption of this function is that the waypoints would be spaced relative to their distance, rather than equaly distributed over time. This would maintain constant traveling speed over the whole path which is something we usually want. But if we would also create a picewiseLerp for Double values instead of Vector2, it's not clear that this would be the desired behaviour.

davesmith00000:
For context: The discussion here was about navigating characters / NPCs around a game level, and the fact that we don't have a nice way to define a path of waypoints that a character will follow.

The above description is quite correct, but additionally I think it would be nice to model this as a `WayPoint` type somehow, so that we can (perhaps later as a second step) do things like emit events when waypoints are reached and have different types of connecting paths and so on.

davesmith00000:
Work completed by @JPonte, will be in the next release.

PR #865, opened 2025-03-16
Add WaypointPath trait

JPonte:
Issue #804

Simply create a path like:
```scala
val path = WaypointPath(
    waypoints = List(Vector2(20, 20), Vector2(80, 30), Vector2(120, 100)),
    radius = 0.0,
    loop = true
)
```

and in either the updateModel or present methods one can get the expected position with:
```scala
path.calculatePosition(at = 0.5)
```
where `at` is the relative position in the path. So 0.0 is the beginning of the path and 1.0 is the end. If looping is enabled values above 1.0 wrap around and values lower than 0.0 also wrap around but in reverse. 

It can be easily integrated in timelines like:
```scala
def loopedPathSignalFunction(waypoints: Vector2*): SignalFunction[Double, (Vector2, Radians)] =
  val path = WaypointPath(waypoints.toList, 0.0, true)
  SignalFunction(over => path.calculatePosition(over))

val traverseSimplePath = loopedPathSignalFunction(Vector2(0, 0), Vector2(1, 0), Vector2(1, 1))

val tl: Timeline[Graphic[Material.ImageEffects]] =
  timeline(
    layer(
      animate(2.seconds) { g =>
        lerp >>> traverseSimplePath >>> SignalFunction((v, r) => g.moveTo(v.toPoint).rotateTo(r))
      }
    )
  )
```


indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
...
+        case (acc, v2) =>
+          acc.lastOption match
+            case Some(v1) =>
+              val fullDistance   = v1.distanceTo(v2)
+              val cappedDistance = (fullDistance - radius).max(0.0)
+              val newV2          = lerpVector2(v1, v2, cappedDistance / fullDistance)
+              acc :+ newV2
+            case None => List(v2)
+    if loop then distances.tail.dropRight(1).prepended(distances.last)
+    else distances
```
JPonte, line 113: I'm not sure how useful this is. The idea is that if the waypoints aren't precise locations but general positions and when a character with a large sprite gets close enough to it, it considers it visited.

JPonte:
@davesmith00000 not sure this is what you had in mind but added a sandbox example of it working with timelines.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
+package indigoextras.waypoints
+
+import indigo.Radians
+import indigo.Vector2
+
+import scala.annotation.tailrec
+
+trait WaypointPath:
+  def waypoints: List[Vector2]
```
davesmith00000, line 9: I shall forever wonder if this is the right thing to have done, but for consistency, `Vector2` should probably be `Vertex`. They're almost identical, but a vertex is a position on a graph / space where vector2 is an expression of direction. `LineSegment` uses `Vertex`, I am open to counter arguments. 😄 

https://github.com/PurpleKingdomGames/indigo/blob/684df20105592b7e9e2d34529621824d584c366b/indigo/indigo/src/main/scala/indigo/shared/geometry/LineSegment.scala#L5

davesmith00000: Additionally, can we replace `List` with `Batch`? `Batch` is _much_ faster and used everywhere.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
...
+  def calculatePosition(
+      at: Double
+  ): (Vector2, Radians)
+
+object WaypointPath:
+  def apply(
+      waypoints: List[Vector2],
+      radius: Double,
+      loop: Boolean
+  ): WaypointPath =
```
davesmith00000, line 24: To keep with Indigo's API style, we could do with adding two more apply methods that call this one:

```scala
def apply(waypoints: List[Vector2]): WaypointPath
def apply(waypoints: List[Vector2], radius: Double): WaypointPath
```

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
+package indigoextras.waypoints
+
+import indigo.Radians
+import indigo.Vector2
+
+import scala.annotation.tailrec
+
+trait WaypointPath:
```
davesmith00000, line 8: Usually I use case classes here to make this easier, but either way we'll probably want method like `def withRadius` that, combined with the comment below about apply methods, allow you to say things like `WaypointPath(v1, v2, v3).withRadius(12)`.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
...
+  def loop: Boolean
+
+  def calculatePosition(
+      at: Double
+  ): (Vector2, Radians)
+
+object WaypointPath:
+  def apply(
+      waypoints: List[Vector2],
+      radius: Double,
```
davesmith00000, line 22: Is radius and expression of `Radians`, or is that too much?

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
...
+
+import indigo.Radians
+import indigo.Vector2
+
+import scala.annotation.tailrec
+
+trait WaypointPath:
+  def waypoints: List[Vector2]
+
+  def effectiveWaypoints: List[Vector2]
```
davesmith00000, line 11: Possible naming suggestion: `effectiveWaypoints` could be `calculatedWaypoints`?

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
davesmith00000: Also we're going to need some Scaladocs.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
...
+      loop: Boolean
+  ): List[Vector2] =
+    val list =
+      if loop then
+        waypoints
+          .appended(waypoints.head)
+      else waypoints
+
+    val distances = list
+      .foldLeft(List.empty[Vector2]):
```
davesmith00000, line 103: I _think_ this only happens during the initial calculation, if so, no big deal, but be aware that `List.foldLeft` is slow and allocates a lot. Something to consider if I'm wrong and this is a hot path.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
davesmith00000: I don't know how this feels in practice, but generally I think that's a very sensible approach. Waiting for things to land perfectly something is a headache and usually not worth the hassle!

Might be worth making this configurable though? How close is close enough? You have a few 'config' options already, if you add any more, consider creating a config object:

```scala
WaypointPathConfig(pathRadius: Radians, loop: Boolean, requiredProximity: Vector2)
object WaypointPathConfig:
  val initial: WaypointPathConfig = WaypointPathConfig(Radians.zero, false, Vector2.one)
```

`requiredProximity` is probably a poor name, it could also be a function that lets the user supply a proximity test - but that's proably overkill for the first cut.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
```diff
@@ -0,0 +1,113 @@
...
+trait WaypointPath:
+  def waypoints: List[Vector2]
+
+  def effectiveWaypoints: List[Vector2]
+
+  def loop: Boolean
+
+  def calculatePosition(
+      at: Double
+  ): (Vector2, Radians)
```
davesmith00000, line 17: Replace the Tuple return type with a nice little case class with named arguments.

davesmith00000 (commented):
It's great! 🔥 

I've peppered it with comments that are mostly about bringing it inline with the rest of Indigo's style. Up to you if you feel like doing the work, I can always merge and refine afterwards.

Either way, the effort is much appreciated. 🙏

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
JPonte: I'm glad you mention this because I just realised yesterday that there is both `Vertex` and `Vector2` and I wasn't sure what was the difference 😅 I'll change both.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
JPonte: I'm not sure what do you mean by this? Radius is about the "proximty" to the waypoint that the actor needs to be from the waypoint for it to be considered visited. It's a measure of distance, not angle.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
JPonte: I think I started with the trait with idea of making different implementations depending on the underlying interpolator (like linear or bezier or something like that). But it's overcomplicating and I think using a case class is easier indeed. I'll add the builder methods as well.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
JPonte: Yes, should only happen during the initial calculation if there is a radius defined. Although I'm not sure it's clear that the users should not recreate the `WaypointPath` often. Something for the docs I guess.
I can also explore an alternative way of doing this.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
JPonte: I added an example of it in the sandbox that hopefully illustrates how it works.

indigo/indigo-extras/src/main/scala/indigoextras/waypoints/WaypointPath.scala
JPonte: Replace the `foldLeft` with a tail recursive function.

davesmith00000:
Glorious. 10/10 demo. 👏 
image

davesmith00000 (approved):
Looks great. I've reviewed the demo - thanks for doing that. 🙏

PR #877, opened 2025-04-06
Add PlaybackPolicy option to PlaySound

### LINKED ISSUE #875: SFX: Add option to stop previous sound
From Discord:

>  Playing a one off sound is easy, you just fire the play sound event.
One thing that I found can be useful (which might not be supported by Indigo, not sure) is to be able to also manually specify channels for sfxs, so that when you start a sound again, it stops the previous one

For example, if you have a "bling" sound when the player presses up and down on a menu, when the player presses one of those keys in rapid succession, sometimes it sounds better to stop the previous bling earlier instead of letting it accumulate

exoego:
Closes #875

- <del>The term "switch" might be not the best.</del>

## Continue 
https://github.com/user-attachments/assets/ba429461-f28c-4a51-b258-732e85bfbe48 

## 🆕 StopPreviousSame 
https://github.com/user-attachments/assets/c3ac3b51-b803-42f0-8cc9-a9186ae14c2f


indigo/indigo/src/main/scala/indigo/platform/audio/AudioPlayer.scala
```diff
@@ -113,9 +114,29 @@ final class AudioPlayer(context: AudioContextProxy):
...
+  def playSound(assetName: AssetName, volume: Volume, switch: SoundSwitch): Unit =
     findAudioDataByName(assetName).foreach { sound =>
-      setupNodes(sound, volume, loop = false).audioBufferSourceNode.start(0)
+      val node = setupNodes(sound, volume, loop = false)
+      node.audioBufferSourceNode.onended = _ => soundNodes.remove(assetName)
+      node.audioBufferSourceNode.start(0)
+      switch match {
+        case SoundSwitch.StopAll          => stopAllSound()
+        case SoundSwitch.StopPreviousSame => stopSound(assetName)
+        case SoundSwitch.Continue         => // noop
```
davesmith00000, line 137: Can this be `=> ()` please? It's probably equivalent, but I prefer something explicit.

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -947,10 +948,15 @@ object KeyboardEvent {
...
+  * @param switch
+  *   How to handle the previous sounds
   */
-final case class PlaySound(assetName: AssetName, volume: Volume) extends GlobalEvent:
+final case class PlaySound(assetName: AssetName, volume: Volume, switch: SoundSwitch = SoundSwitch.Continue)
+    extends GlobalEvent:
   def withVolume(newVolume: Volume): PlaySound =
     this.copy(volume = newVolume)
+  def withSwitch(newSwitch: SoundSwitch): PlaySound =
+    this.copy(switch = newSwitch)
```
davesmith00000, line 959: Can we remove the default argument, and add something like the below instead? Indigo doesn't employ default arguments anyway. I like to be explicit about valid combinations - which they all are in this case - but it keeps it consistent.

```scala
object PlaySound:

  def apply(assetName: AssetName): PlaySound =
    PlaySound(assetName, Volume.Max, SoundSwitch.Continue)
    
  def apply(assetName: AssetName, volume: Volume): PlaySound =
    PlaySound(assetName, volume, SoundSwitch.Continue)
```

indigo/indigo/src/main/scala/indigo/platform/audio/SoundSwitch.scala
```diff
@@ -0,0 +1,14 @@
...
+    */
+  case StopAll
+
+  /** Stop only the previous same sound.
+    */
+  case StopPreviousSame
+
+  /** Continue all previous sounds
+    */
+  case Continue
```
davesmith00000, line 14: I agree that `SoundSwitch` as a name is a bit clunky. How about something like either `SoundOverlap` or `PlaybackPolicy`?

davesmith00000 (commented):
Lookin' good, I've left a couple of comments though.

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
exoego: Renamed in https://github.com/PurpleKingdomGames/indigo/pull/877/commits/980d8e0093f9abddb244551432259bc33775c7da

indigo/indigo/src/main/scala/indigo/platform/audio/SoundSwitch.scala
exoego: Renamed in https://github.com/PurpleKingdomGames/indigo/pull/877/commits/980d8e0093f9abddb244551432259bc33775c7da

indigo/indigo/src/main/scala/indigo/platform/audio/AudioPlayer.scala
exoego: Addressed in https://github.com/PurpleKingdomGames/indigo/pull/877/commits/5f02245f246299278ac5fc7a6e196e71d9713dff

indigo/indigo/src/main/scala/indigo/shared/events/GlobalEvent.scala
```diff
@@ -947,10 +948,22 @@ object KeyboardEvent {
...
   * @param volume
   *   What volume level to play at
+  * @param policy
+  *   How to handle the previous sounds
   */
-final case class PlaySound(assetName: AssetName, volume: Volume) extends GlobalEvent:
+final case class PlaySound(assetName: AssetName, volume: Volume, policy: PlaybackPolicy) extends GlobalEvent:
   def withVolume(newVolume: Volume): PlaySound =
     this.copy(volume = newVolume)
+  def withSwitch(newPolicy: PlaybackPolicy): PlaySound =
```
davesmith00000, line 957:
```suggestion
  def withPlaybackPolicy(newPolicy: PlaybackPolicy): PlaySound =
```

davesmith00000 (approved):
Looks good, I've suggested a small change (left over from the switch name).

davesmith00000:
I'm trying to sort out a release and I'd like to include this - it's a good change! So I'm going to push a commit that moves a couple of small things around and adds a demo. Won't be long.

davesmith00000 (approved):

PR #878, opened 2025-04-06
Refactor use of Map.put for perf and clarity 

exoego:
This is a minor refactoring 

- for performance, by allocating less `Option` returned by `put` and `remove`
- for clarity, by using less `@nowarn`


davesmith00000 (approved):
Good stuff, some of that was rather old code! Thanks!

PR #879, opened 2025-04-06
Add option to open electron devtool

exoego:
I found electron's devtool is handy when debugging


davesmith00000 (approved):
This is great, thanks. The only thing missing is a couple of helper methods on the `IndigoElectronOptions` type, but I'll add those post-merge.

PR #886, opened 2025-04-09
Fixed #874: Dragging a window does not focus it

### LINKED ISSUE #874: Bug: Dragging a window doesn't focus it

davesmith00000:
This was a tricky fix in the end @ivan-tatsenko-cci! 😉


ivan-tatsenko-cci:
> This was a tricky fix in the end @ivan-tatsenko-cci! 😉

Was it one of those one-line commits that require a couple of hours of investigation? :)

PR #888, opened 2025-04-12
hot-reloading

exoego:
- [x] Relies on [vite-plugin-electron](https://github.com/electron-vite/vite-plugin-electron)
- [ ] Symlink to avoid file copy 

currently, not so impressive
https://github.com/user-attachments/assets/128fea64-bf72-4db7-991d-ac8b5fcedaf1


davesmith00000:
> currently, not so hot

🤣 

I've only just got back from being away, will take a closer look soon. Interestingly, this isn't what I had in mind... bit it's interesting!

I was thinking more along the lines of the hot reload module in Tyrian, which doesn't have an electron integration. But I look forward to reading this properly. 😄

PR #904, opened 2025-05-04
Layer rendering performance tuning

davesmith00000:


davesmith00000:
This PR adds a few small optimisations to reduce the number of graphics card calls we make - and that did help a bit - but the real performance boost has been made by introducing 'layer compacting'.

The idea is simple: Drawing layers is slow, and the process is complicated, so let's just draw less of them. Every top level `LayerEntry` is preserved, and layer stacks are flattened as they aways used to be, but now content layer batches (under a key) are squashed into the previous one if they have the properties, i.e. if every layer is unique, every layer will be preserved and rendered, but if all the layers had the same properties, you'd end up with a single layer to render.

In the case of UI rendering, this means _many_ layers get flattened into just a few.

From (with some optimisations):

image

To (with layer compacting):

image

PR #931, opened 2025-06-18
Allows buttons to have a hover state

### LINKED ISSUE #928: Buttons need an OnEnter and OnExit
It's sometimes useful to have an audio cue when a button is hovered over. The only way to achieve this at the moment is by adding an event on the presentation outcome, which causes an audio cue to fire indefinitely until the pointer exits the button.

It would be better to listen for the pointer entering the button, and then fire an event once which triggers an audio cue.

hobnob:
`Buttons` and `HitArea`s now have a state determining if the pointer has entered or left the component. This allows the component to fire events such as audio cues once a pointer has entered a button.

Closes #928


davesmith00000 (approved):
Very neat and tidy - looks good!

PR #936, opened 2025-06-24
Improving Input Handling

hobnob:
The state of inputs in Indigo is being updated to account both for the broad range of input devices and to simplify the generic pointer events. Previously the Pointer events have been largely carbon copies of the JS equivalents, but this doesn't necessarily track with the rest of Indigo, which has been largely platform agnostic.

The plan here is to provide a much-simplified set of `Pointer` events that provide access to position and button data. These can be used for UI components or other parts of a game where 'how' a player interacted with something is unimportant, meaning that a `Click` event could mean from a `Mouse`, `Touch`, or `Pen`.

Because a `Wheel` input can be attached to any device, not just a `Mouse`, the `MouseWheel` event will be replaced with 4 new events: 

 * `WheelEvent.Move` - for any movement at all on a wheel input
 * `WheelEvent.Horizontal` - for any movement on the horizontal axis of a wheel
 * `WheelEvent.Vertical` - for any movement on the vertical axis of a wheel
 * `WheelEvent.Depth` - for any movement on the Z axis of a wheel., This is more conceptual than anything else, but is here for future proofing

New events will be added to account for `Pen` and `Touch`, allowing for greater control over those areas where a player might interact in a specific way with those inputs (for example pinch-zooming)

The `InputState` will be updated to deprecate the `InputState.mouse.scrolled` and 3 new properties have been added to `InputState`:

* `touch` - For all touch related states
* `pen` - For all pen related input states
* `wheel` - For the current state of the wheel input


davesmith00000 (approved):
Looks great! I'll catch up with you on Slack just to confirm you're happy to merge. Sorry for the slow review! 😁

PR #939, opened 2025-06-25
Adds `withMasterVolume` to `SceneAudio`

### LINKED ISSUE #938: Provide a nice way of setting the master volume on `SceneAudio`
When generating assets throught the generator scripts, audio is represented 3 times, one of which is as `SceneAudio`, which can be used for playing looping audio such as background music.

Changing the volume for this audio, however, means copying the `SceneAudio` and manipulating any sources within it to have the desired master volume. This feels a bit clunky, and so I propose either a `withMasterVolume` helper method on `SceneAudio`, or an optional second parameter on `withAudio` that is the master volume.

hobnob:
Closes #938

### DIFF
--- a/indigo/indigo/src/main/scala/indigo/shared/scenegraph/SceneAudio.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/scenegraph/SceneAudio.scala
@@ -15,9 +15,9 @@ final case class SceneAudio(
 
   def withMasterVolume(volume: Volume): SceneAudio =
     SceneAudio(
-      sourceA.map(s => s.copy(masterVolume = s.volume * volume)),
-      sourceB.map(s => s.copy(masterVolume = s.volume * volume)),
-      sourceC.map(s => s.copy(masterVolume = s.volume * volume))
+      sourceA.map(s => s.copy(masterVolume = volume)),
+      sourceB.map(s => s.copy(masterVolume = volume)),
+      sourceC.map(s => s.copy(masterVolume = volume))
     )
 
 object SceneAudio:
--- a/indigo/indigo/src/main/scala/indigo/shared/scenegraph/SceneAudio.scala
+++ b/indigo/indigo/src/main/scala/indigo/shared/scenegraph/SceneAudio.scala
@@ -6,6 +6,7 @@ import indigo.shared.audio.Volume
   * audio sources at once.
   */
 final case class SceneAudio(
+    masterVolume: Volume,
     sourceA: Option[SceneAudioSource],
     sourceB: Option[SceneAudioSource],
     sourceC: Option[SceneAudioSource]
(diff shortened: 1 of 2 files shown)


indigo/indigo/src/main/scala/indigo/shared/scenegraph/SceneAudio.scala
```diff
@@ -11,6 +13,13 @@ final case class SceneAudio(
   def |+|(other: SceneAudio): SceneAudio =
     SceneAudio.combine(this, other)
 
+  def withMasterVolume(volume: Volume): SceneAudio =
+    SceneAudio(
+      sourceA.map(s => s.copy(masterVolume = s.volume * volume)),
+      sourceB.map(s => s.copy(masterVolume = s.volume * volume)),
+      sourceC.map(s => s.copy(masterVolume = s.volume * volume))
+    )
```
davesmith00000, line 21: I don't think that quite works, does it? What you're doing that is affecting the volume of all audio sources - which is ok - but I don't think that's the same as a master volume setting.

Two solutions:

1. Keep what you've got and change the name from `withMasterVolume` to something like `changeValueBy` to imply that you're affecting the relative volume of all tracks, on the audio sources.
2. Add a new field `masterVolume` to `SceneAudio, change `withMasterVolume` to just set that value, and have the audio sources somehow refer to it when choosing their own volume. This means that from a data perspective, `SceneAudio` holds the 'master / global volume' setting, and whatever the volume in each `SceneAudioSource` instance is (weirdly also called masterVolume, maybe rename?), is relative to the top level master volume.

The difference between 1 and 2 is that 1 is destructive and 2 isn't. If you call `withMasterVolume` in (1) with a value of 0, then all sources are set to 0 and you lose the relative volumes of the other three. In (2), setting `withMasterVolume` to 0 outputs silence, but the audio sources actually keep their relative volumes so that when you dial it up again, they stay relative. If that makes sense.

hobnob: Yes, you're right it should be destructive.. I've changed it

davesmith00000: Just to capture our slack chat:

There's three ways we could globally adjust the volume:

1. overrideVolume - all volumes are now this much (destructive)
2. changeVolumeBy - Relatively turn all the dials by this much, but if they all hit zero, turning them up again will mean they're level (destructive)
3. 'master volume' - There's a new volume value in the SceneAudio, and all source audio's are multiplied by that, preserving relative volume settings that the user has carefully chosen. (non-destructive).

I quite like them all, I think they have different use cases.

I also think that `masterVolume` on the `SceneAudioSources` is misnamed, and should just be `volume`.

Issue #949, opened 2025-08-03
mill-indigo for Mill 1.x.y


susliko:
Hi 👋🏻 

At the moment `mill-indigo` plugin relies on Mill `0.12.x`. 
According to the [1.0.0 release notes](https://github.com/com-lihaoyi/mill/blob/main/changelog.adoc#100) all the plugins need to be re-published  to be used with new major version of Mill

davesmith00000:
Hi, thanks for raising!

I'm currently doing some work on Tyrian, as soon as I'm done I'll be moving on to Indigo and I'll pick this up (assuming no one beats me to it).

davesmith00000:
Hello again, great, I'll take a look at the PR!

Honestly we usually just force migrate people to the latest version of everything. If we had a team of people working on this we could afford to worry about backwards compatibility... but ya know... we don't. 😆

Having said that, if the migration is significant, maybe we could try and do something to support both - but for now - carry on, I think you've made the right decision.

davesmith00000:
As a matter of fact, I was wondering if forcing people up to sbt 2.0 _now_ was an idea worth pursuing. Then everything is just Scala 3. But I haven't looked into it at all.

PR #950, opened 2025-08-11
Cross-builds indigo-plugin module for Scala 3

susliko:
Indigo provides plugins for both sbt and mill. Both plugins have a common `indigo-plugin` dependency that allows avoiding code duplication. For now it's cross-built for Scala 2.12 (sbt) and Scala 2.13 (mill 0.12.x). This PR adds cross-build for Scala 3.3 LTS (which is a preferred version for libraries) to support mill 1.0 migration https://github.com/PurpleKingdomGames/indigo/issues/949

### DIFF
--- a/indigo-plugin/build.sc
+++ b/indigo-plugin/build.sc
@@ -9,7 +9,7 @@ import mill.scalalib.scalafmt._
 import coursier.maven.MavenRepository
 import publish._
 
-object `indigo-plugin` extends Cross[IndigoPluginModule]("2.12", "2.13")
+object `indigo-plugin` extends Cross[IndigoPluginModule]("2.12", "2.13", "3")
 
 trait IndigoPluginModule extends CrossScalaModule with PublishModule with ScalafmtModule {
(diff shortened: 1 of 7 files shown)


indigo-plugin/build.sc
```diff
@@ -19,6 +19,7 @@ trait IndigoPluginModule extends CrossScalaModule with PublishModule with Scalaf
     crossScalaVersion match {
       case "2.12" => "2.12.20"
       case "2.13" => "2.13.16"
+      case "3"    => "3.3.6" // Relying on LTS Scala
```
davesmith00000, line 22: LTS seems like a sensible move, and yet:

> build.mill and package.mill files are now written in Scala 3.7.1

https://github.com/com-lihaoyi/mill/blob/main/changelog.adoc#major-breaking-changes

🤷 

I've asked if there's a recommendation:
https://discord.com/channels/632150470000902164/1295786923410984981/1404585711981691012

